PR #4, opened 2015-12-04
Add README and RELEASE_NOTES

andreas-bok-sociomantic:
Add README File

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,77 @@
+Description
+===========
+
+Drizzle Redux is a project which aims to breath new life into the libdrizzle
+C connector. It is designed to allow you to connect to and query a
+MySQL database server using a simple API.
+
+Libdrizzle-redux in the sociomantic repository is a fork of the official
+libdrizzle redux at https://launchpad.net/libdrizzle.
+
+Releases
+========
+
+`Latest stable release notes
+<https://github.com/sociomantic/libdrizzle-redux/releases/latest>`_ | `Current,
+in development, release notes
+<https://github.com/sociomantic/libdrizzle-redux/blob/master/RELEASE_NOTES.md>`_ | `All
+releases <https://github.com/sociomantic/libdrizzle-redux/releases>`_
+
+Releases are handled using `GitHub releases
+<https://github.com/sociomantic/libdrizzle-redux/releases>`_. The release notes
+provided there are usually structured in 3 sections, a **Migration Instructions**,
+which are the mandatory steps the users have to do to update to a new version,
+**Deprecated** which contains deprecated functions which is recommended not to
+use but will not break any old code, and the **New Features** which are
+optional new features available in the new version that users might find
+interesting.  Using them is optional, but encouraged.
+
+These instructions should help developers to migrate from one version to
+another. The changes listed here are the steps you need to take to move from
+the previous version to the one being listed. For example, all the steps
+described in version **v1.5** are the steps required to move from **v1.4** to
+**v1.5**.
+
+If you need to jump several versions at once, you should read all the steps from
+all the intermediate versions. For example, to jump from **v1.2** to **v1.5**,
+you need to first follow the steps in version **v1.3**, then the steps in
+version **v1.4** and finally the steps in version **v1.5**.
+
+There are also sometimes *patch-level* releases, in that case there are no
+breaking changes or new features, just bug fixes, and thus, only bug fixes are
+listed in the release notes.
+
+Building libdrizzle-redux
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -0,0 +1,117 @@
...
+-  Currently missing server-side prepared statement support and protocol compression
+
+
+Dependencies
+^^^^^^^^^^^^
+
+:zlib: v1.x
+:openssl or ssl: v1.x
+
+Migration Instructions
```
ben-palmer-sociomantic: I don't think you need the migration instructions here. They can live in the release notes.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,42 @@
...
+Dependencies
+------------
+
+Dependency | Version
+-----------|---------
+zlib           | v1.x
+openssl or ssl | v1.x
+
+
+Migration Instructions
```
ben-palmer-sociomantic: If there are no migration instructions I would just leave this section out.

ben-palmer-sociomantic:
I personally am fine with having the build instructions in the README rather than using a wiki page.

RELEASE_NOTES.md
andreas-bok-sociomantic: The **TODO** list is for Migration Instructions. 
But it might be obsolete if we assume users are starting with `v5.1.4`. 

Might be more relevant with Implementation examples, eg. the examples from the pdf manual

ben-palmer-sociomantic: > But it might be obsolete if we assume users are starting with v5.1.4

I think that is a reasonable assumption to make and then we maintain migration instructions from there onwards. @leandro-lucarella-sociomantic do you agree?

README.rst
```diff
@@ -0,0 +1,105 @@
+Description |BuildStatus|_
```
leandro-lucarella-sociomantic: The `|BuildStatus|_` is wrong as it's not pointing at anything. We should probably build and test this project using travis-ci.org for example.

andreas-bok-sociomantic:
Would it be more appropriate with an `INSTALL` file with build/installation instructions?

README.rst
andreas-bok-sociomantic: @leandro-lucarella-sociomantic, Should I look into setting up build and test on `travis-ci.org` or what you suggest? 
Any case I could make an issue for this.

andreas-bok-sociomantic: Created an issue #6 for CI

andreas-bok-sociomantic:
- updated with changes from comments 
- squashed commits

ben-palmer-sociomantic:
LGTM, if you have nothing else @leandro-lucarella-sociomantic I am happy to merge.

ben-palmer-sociomantic:
Actually, I just noticed the commit message should probably have the "Not sure where to put...." removed.

andreas-bok-sociomantic:
Updated commit msg.

andreas-bok-sociomantic:
I managed to build the pdf manual: `make pdflatex`
It did require fixing quite a lot of missing latex packages. 

The generated pdf is the same as the previously uploaded pdf. 
To me it therefore makes more sense to remove this file.

leandro-lucarella-sociomantic:
Having a release notes for a version we didn't release seems to be a bit weird, specially because they should be cleared right in the next commit, but I'm not strongly against this.

Another option would be to just tag the imported repo as `v5.1.4` and put the release notes directly in the github release notes. That might make more sense.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,43 @@
+# v5.1.4 libdrizzle redux (2013-04-25)
+
+# Changelog
+
```
leandro-lucarella-sociomantic: BTW, the data above shouldn't be part of the release notes, as the version will be present in the github release (remember this file is just used to collect notes that will be put in the github release page) and the changelog title is redundant :)

ben-palmer-sociomantic:
> Another option would be to just tag the imported repo as v5.1.4 and put the release notes directly in the github release notes. That might make more sense.

I think I would be in favour of this option as well actually.

andreas-bok-sociomantic:
Removed RELEASE_NOTES.md file and added a GITHUB release note instead

ben-palmer-sociomantic:
LGTM, if you squash the commits I will rebase this.

ben-palmer-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was a972accf19fb88b7afaa8ab98cbf6bb9db18d52d, new (rebased) HEAD is 218e98d9d427c004c87da4c47e771593b58b2047

Issue #7, opened 2015-12-08
Modify unittests 


andreas-bok-sociomantic:
Currently `unittests` depends on environment variables to set up connection to `mysql` db

```
tests/unit/binlog.cc:73 - main()

drizzle_st *con= drizzle_create(getenv("MYSQL_SERVER"),
    getenv("MYSQL_PORT") ? atoi("MYSQL_PORT") : DRIZZLE_DEFAULT_TCP_PORT,
    getenv("MYSQL_USER"),
    getenv("MYSQL_PASSWORD"),
    getenv("MYSQL_SCHEMA"), 0);
```

Only `MYSQL_PORT` has a fallback value if it is not defined. Hence if `MYSQL_USER, MYSQL_PASSWORD, MYSQL_SCHEMA` defined the tests will fail. 

Possible solutions
- Keep the dependency on environment variables. But info about `env var` dependency
- Move mysql config to a settings file in the `test/` folder
- Add default values for connection parameters in test source code. 

What are your thoughts, @leandro-lucarella-sociomantic , @ben-palmer-sociomantic

ben-palmer-sociomantic:
For `MYSQL_SERVER` it makes sense to me to use `127.0.0.1` but for user, password, and schema I don't see obvious default values. Perhaps a settings file in the test folder would be the best way to go.

leandro-lucarella-sociomantic:
Yeah, I think defaults in here will make very little sense. I don't mind keeping the env vars. Using a config file could be also nice, but bear in mind that file shouldn't be part of the repo, as it's local configuration. The config file can be simply a trivial shell script defining the variables:

``` sh
MYSQL_SERVER=blah
MYSQL_PORT=1
# etc.
```

then `. config.sh; run-tests`.

leandro-lucarella-sociomantic:
In any case, it should be well documented, and ideally the test should check all the env vars are defined and give a sensible error (asking to set the variables) if it's not.

Another option, if it's an option, would be to pass the mysql through the command line to the tests. `run-tests localhost user pass schema [port]`.

andreas-bok-sociomantic:
A function exists in https://github.com/sociomantic/libdrizzle-redux/blob/master/tests/unit/common.c#L76 for creating a schema. 
Maybe a function could be added to set up a `test_user` as well, so `db config`  is independent of the `env` variables. 
But could run into errors due `privileges`...or

andreas-bok-sociomantic:
[...]

Issue #6, opened 2015-12-08
Add continuous integration


andreas-bok-sociomantic:
Continuous integration for the repo should be added

A list of Continous Integration Services is here: 
https://github.com/ligurio/Continuous-Integration-services/blob/master/continuous-integration-services-list.md

Free for opensourcea and with database support: 
[TravisCI](https://travis-ci.org/) 
[SemaphoreCI](https://semaphoreci.com/docs/)

leandro-lucarella-sociomantic:
I've used travis-ci before, and it's OK. You just have to include a `.travis.yml` (or something like that) in the project with the commands you want to run and that's it. Oh! And set up the hooks.

andreas-bok-sociomantic:
Ok, I will go for `travis-ci`. 
Will require a minor change in `bootstrap.sh` since `jenkins` was used for CI in original branch. 
Should the check for the `jenkins` support simply be removed?

andreas-bok-sociomantic:
Should I create an account on travis-ci.org or do we have **sociomantic** account for tsunami projects already?

andreas-bok-sociomantic:
I can't add the repo to travis-ci, as I am not admin on the repo. @leandro-lucarella-sociomantic  Is it better that you add the repo to travis and activate it?

leandro-lucarella-sociomantic:
I guess it would be better for @ben-palmer-sociomantic to do it. But another option would be to make you admin of this repo, as you'll be the one actually doing most of the work. I leave that to Ben.

ben-palmer-sociomantic:
![screenshot from 2015-12-10 11 52 07](https://cloud.githubusercontent.com/assets/2894657/11713605/bf75fe16-9f34-11e5-8cdd-cbf12f663a66.png)

ben-palmer-sociomantic:
@leandro-lucarella-sociomantic this are the requested permissions for the travis ci testing. Is this ok to authorise? I don't see any problems but seeing the organization access made me nervous and I wanted to double check.

ben-palmer-sociomantic:
So after briefly granting access to travis-cl to this repo from my sociomantic account it seems this needs organization access so it would be better to do it from a non sociomantic account so I have reverted the application permissions. See also https://github.com/sociomantic/microbe/issues/67 for microbe. So instead I will grant administrator rights to @bokchan and he can grant the permissions to the application without involving the organization.

ben-palmer-sociomantic:
[...]

PR #11, opened 2015-12-15
Add travis-ci config file

andreas-bok-sociomantic:
Currently test db setup depends on ENVIRON VARS

Running tests, thus requires #7 to be resolved

Addresses #6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,56 @@
+language : cpp
+
+sudo: required
+
+dist: trusty
+
+services:
+  - mysql
+
+## The following environment variables are required for 
+#  running unittests 
+#  Environment variables can be defined in the repository, but 
+#  are overridden by values in .travis.yml. 
+#  
+#  Please refer to the section 'script' for info on 
+#  SKIP_TEST_ON_ERROR
+env: 
+  - MYSQL_SERVER=127.0.0.1
+  - MYSQL_PASSWORD=
+  - MYSQL_USER=root
+  - SKIP_TEST_ON_ERROR=0 
+
+## Defines packages to be installed by apt
+#  Version is specified since some tests require MySQL >=5.6 
+addons:
+  apt:
+    packages:
+    - mysql-client-core-5.6 
+    - mysql-server-core-5.6
+    - mysql-client-5.6
+    - mysql-server-5.6
+
+## This script runs before the compiling
+#  1. Stop MySQL server
+#  2. Enable replication in mysql config file
+#  3. Start MySQL server
+#  4. Print mysql config for debug purposes
+before_script:
+  - sudo service mysql stop
+  - sudo sed  -i 's/\[mysqld\]/[mysqld]\nserver-id = 1\nlog-bin = mysqlbin\nbinlog-format = ROW\n/g' /etc/mysql/my.cnf
+  - sudo service mysql start 
+  - mysqld --print-defaults 
+
+## Build steps 
+#  - compile project 
+#  - run all tests with 'make check'
+# 
+#  Please note that test fails on first error as default
+#  To skip errors and fail silently set environment variable: 
+#    SKIP_TEST_ON_ERROR=1
+script:
+  - autoreconf -fi
+  - ./configure
+  - make
+  - make check
+
--- a/yatl/lite.h
+++ b/yatl/lite.h
@@ -52,8 +52,14 @@
 # define __PRETTY_FUNCTION__ __func__
 #endif
 
+#ifndef SKIP_TEST_ON_ERROR
+ #define SKIP_TEST_ON_ERROR (getenv("SKIP_TEST_ON_ERROR") ? \
+                             atoi(getenv("SKIP_TEST_ON_ERROR")) : \
+                             0)
+#endif
+
 #ifndef EXIT_SKIP
-# define EXIT_SKIP 77
+ #define EXIT_SKIP SKIP_TEST_ON_ERROR ? 77 : EXIT_FAILURE
 #endif
 
 static inline bool valgrind_is_caller(void)


andreas-bok-sociomantic:
Adding travis config is a bit of a trial and error process. 
Installing dependencies on in the container running the tests require installing a package, `libcloog-ppl0`, which is currently not on the travis whitelist. 
Created an issue for adding this at https://github.com/travis-ci/apt-package-whitelist/issues/2134

andreas-bok-sociomantic:
The configuration is atm targeted at 
`ubuntu 14.04 trusty`. 

Issues were related to testing on travis precise: 
1. `automake` was `v1.11`, doesn't support `[serial-test]` as 
test-harness
2. An compile error which was AFAICS caused by the provided GCC v4.6.2 having been installed without support for Graphite loop optimizations. 

> libdrizzle/query.cc:149:6: sorry, unimplemented: Graphite loop optimizations can only be used if the libcloog-ppl0 package is installed

andreas-bok-sociomantic:
The project compiles and passes tests with the configuration defined in `.travis.yml`. 

`binlog` tests are ignored since it requires `mysql` to be started with `--log-bin=path-to-binlog-file`

ben-palmer-sociomantic:
LGTM, maybe squash the commits and then I can rebase

andreas-bok-sociomantic:
I just wanna see if I can set up mysql with binlog enabled before rebase

ben-palmer-sociomantic:
> I just wanna see if I can set up mysql with binlog enabled before rebase

Ok, no problem. Let me know when you want me to have another look at it.

andreas-bok-sociomantic:
The travis configuration now compiles and passes all tests 
Rebased on upstream/master.
Ready for review

ben-palmer-sociomantic:
I am happy with this method of doing the tests. @leandro-lucarella-sociomantic any comments?

leandro-lucarella-sociomantic:
Yes: Niiiice!

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic do you have another suggestion for changing the mysql config. The `sed` inplace substitute is not the most elegant/fail-safe solution I reckon...

leandro-lucarella-sociomantic:
Not really. Is not ideal but I think it's good enough. Will you find out if the file changes, and the `sed` command does nothing? If that's the case, then is a safe hack. If not, we might want to add some check to make sure we are either really testing this, or the tests get totally broken. What we need to avoid is silent breaking and suddenly thinking all is good and nothing is being properly tested.

andreas-bok-sociomantic:
> Will you find out if the file changes, and the sed command does nothing?

Made a `diff` against a copy before `sed` of `my.cnf`. So the `sed` does change the mysql config 

``` sh
$ sudo service mysql stop
mysql stop/waiting
before_script.2
$ sudo cp /etc/mysql/my.cnf /etc/mysql/my.bck
before_script.3
$ sudo sed  -i 's/\[mysqld\]/[mysqld]\nserver-id = 1\nlog-bin = mysqlbin\nbinlog-format = ROW\n/g' /etc/mysql/my.cnf
$ diff /etc/mysql/my.cnf /etc/mysql/my.bck
32,35d31
< server-id = 1
< log-bin = mysqlbin
< binlog-format = ROW
```

>  What we need to avoid is silent breaking and suddenly thinking all is good and nothing is being properly tested.

Tests fail silently if the error is due to missing feature support in build environment. 
@leandro-lucarella-sociomantic Given this behaviour, would you recommend breaking the tests on first error by changing the source of the test cases?

```
tests/unit/binlog.cc:81: main SKIP 'ret == DRIZZLE_RETURN_ERROR_CODE' [ Binlog is not open?: Binary log is not open(DRIZZLE_RETURN_ERROR_CODE) ]
SKIP: tests/unit/binlog
PASS: tests/unit/escape
PASS: tests/unit/insert_id
PASS: tests/unit/query
PASS: tests/unit/column
...
tests/unit/datetypes.c:90: main SKIP 'drizzle_server_version_number(con) < 50604' [ Test requires MySQL 5.6.4 or higher ]
SKIP: tests/unit/datetypes
...
======================
All 14 tests passed
(2 tests were not run)
```

ben-palmer-sociomantic:
>  Given this behaviour, would you recommend breaking the tests on first error by changing the source of the test cases?

That seems like a good idea to me. If the tests need to be skipped then they should fail.

andreas-bok-sociomantic:
Whether tests should fail on first error or silently can now be set via env var 
`SKIP_TEST_ON_ERROR`
Setting the var is optional as it evaluates to false if it is **not** defined.

ben-palmer-sociomantic:
> Setting the var is optional as it evaluates to false if it is defined.

Should this be "if it is NOT defined"?

andreas-bok-sociomantic:
Yes, typo.

ben-palmer-sociomantic:
All the whitespace changes should be in their own commit then LGTM

andreas-bok-sociomantic:
Removed whitespace changes caused by Sublime's, otherwise, helpful trailing whitespace trimming

ben-palmer-sociomantic:
ok, LGTM

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic any comments?

leandro-lucarella-sociomantic:
> Made a diff against a copy before sed of my.cnf. So the sed does change the mysql config

Sorry, what I meant was what would happen if the config file in travis changes and suddenly the sed pattern stop working so it does nothing. Would we detect that problem via failing tests?

leandro-lucarella-sociomantic:
> Tests fail silently if the error is due to missing feature support in build environment.
> @leandro-lucarella-sociomantic Given this behaviour, would you recommend breaking the tests on first error by changing the source of the test cases?

Yes, I think that's a good idea. It might be more annoying for users wanting to test, since they'll need to have a complete environment to test, but I guess is a fair thing to ask, and worse case you can just use travis for a complete test.

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+
+sudo: required
+
+dist: trusty
+
+services:
+  - mysql
+
+## The following environment variables are specified in 
+#  the repository settings at travis-ci.org
```
leandro-lucarella-sociomantic: Any reason why not to specify them here, so any developer can modify them via a PR?

yatl/lite.h
```diff
@@ -52,8 +52,14 @@
...
 
+#ifndef SKIP_TEST_ON_ERROR
+ #define SKIP_TEST_ON_ERROR (getenv("SKIP_TEST_ON_ERROR") ? \
+                             atoi(getenv("SKIP_TEST_ON_ERROR")) : \
+                             0)
+#endif
+
 #ifndef EXIT_SKIP
-# define EXIT_SKIP 77
+ #define EXIT_SKIP SKIP_TEST_ON_ERROR ? 77 : EXIT_FAILURE
```
leandro-lucarella-sociomantic, line 62: Wait, if some tests are skipped then the test program exits with a code 77 and not 0? If so the tests should fail already!

andreas-bok-sociomantic:
> Sorry, what I meant was what would happen if the config file in travis changes and suddenly the sed pattern stop working so it does nothing. Would we detect that problem via failing tests?

We would get an error iff `SKIP_TEST_ON_ERROR=1`. The binlog test depends on MySQL having replication enabled. As of now the `sed` substitution  in `my.cnf` is doing that.
However we should accommodate the script ifthe format of the mysql config file changes in the future.

andreas-bok-sociomantic:
> Any reason why not to specify them here, so any developer can modify them via a PR?

Admittedly I did exactly to avoid doing a PR when testing the env variables. However defining the variables in the `.travis.yml` file overrules the values defined in the settings at `travis-ci.org`
So I will uncomment the vars in the config file.

andreas-bok-sociomantic:
> Wait, if some tests are skipped then the test program exits with a code 77 and not 0? If so the tests 
> should fail already!

Yes. but the unittests are separate programs so the unittest report AFAICS simply gathers the return code of the individual tests. 

However with `SKIP_TEST_ON_ERROR=0` skipped tests return `EXIT_FAILURE`. So the travis build will show build as failed in this case

bokchan:
test ping @andreas-bok-sociomantic

bokchan:
@andreas-bok-sociomantic

andreas-bok-sociomantic:
Added environment variables to `.travis.yml` file

leandro-lucarella-sociomantic:
> Yes. but the unittests are separate programs so the unittest report AFAICS simply gathers the return code of the individual tests.
> 
> However with SKIP_TEST_ON_ERROR=0 skipped tests return EXIT_FAILURE. So the travis build will show build as failed in this case

Let me see if I understand this correctly, so if the program returns 77, so test runner picks up that return code as skipped and won't fail the overall test?

andreas-bok-sociomantic:
> Let me see if I understand this correctly, so if the program returns 77, so test runner picks up that return code as skipped and won't fail the overall test?

As such we could hard code `EXIT_SKIP` as 1 and any skipped test would cause the overall test to fail. However I wanted to make it possible to decide skipped tests should cause the overall test should fail. Hence the env var `SKIP_TEST_ON_ERROR`

andreas-bok-sociomantic:
Squashed commits. Should be ready for merge if there's no further comments

ben-palmer-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was 6bfd9c6239e4642599f5424c4c70d303fdfb0a70, new (rebased) HEAD is 4df8e0679a93e3ffc2959cef636395d8126e04f1

PR #14, opened 2016-01-08
Add deb packaging support

andreas-bok-sociomantic:
Makefile now offers a target for **deb packaging**
The `deb` package is built using `fpm`:

```
https://github.com/jordansissel/fpm
```

Added `automake` include file for `deb`

Added script file, `deb/build` to build package using `fpm`

Add `/deb/include.am` to top level `Makefile.am`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,6 +41,7 @@ include m4/include.am
 include libdrizzle/include.am
 include libdrizzle-5.1/include.am
 include rpm/include.am
+include deb/include.am
 include docs/include.am
 
 EXTRA_DIST+= AUTHORS
@@ -69,5 +70,6 @@ maintainer-clean-local:
 	find . -type f -name '*~' -exec rm -f '{}' \;
 	-rm -f @PACKAGE@-*.tar.gz
 	-rm -f @PACKAGE@-*.rpm
+	-rm -f @PACKAGE@-*.deb
 
 dist_bin_SCRIPTS+= @GENERIC_CONFIG@ 
--- a/deb/build
+++ b/deb/build
@@ -1,0 +1,42 @@
+#!/bin/sh
+set -e
+
+cd `dirname $0`
+
+genchangelog()
+{
+        echo "$1 ($2) `lsb_release -sc`; urgency=low"
+        echo
+	prevtag=$(git describe --abbrev=0 HEAD^)
+        git log --date=short --format="  * %s (%h, %cd)" "$prevtag"..HEAD |
+                fold --spaces --width 76 | sed 's/^\([^ ]\+\)/    \1/'
+        echo
+        echo " -- $3  `LANG=C date -R`"
+}
+
+pkgversion=$(git describe --dirty | cut -c2- |
+		sed 's/-\([0-9]\+\)-\(g[0-9a-f]\+\)/+\1~\2/' |
+		sed 's/\(~g[0-9a-f]\+\)-dirty$/-dirty\1/' |
+		sed 's/-dirty/~dirty.'`date +%Y%m%d%H%M%S`'/'
+	)-$(lsb_release -cs)
+
+pkgmaint=$(echo "`git config user.name` <`git config user.email`>")
+
+changelog=`mktemp`
+trap "rm -f '$changelog'; exit 1" INT TERM QUIT
+
+pkgname=libdrizzle-redux
+genchangelog "$pkgname" "$pkgversion" "$pkgmaint" > "$changelog"
+fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
+	--architecture all \
+	--maintainer "$pkgmaint" \
+	--description "Simplified API to MySQL databases" \
+	--url 'https://github.com/sociomantic/libdrizzle-redux' \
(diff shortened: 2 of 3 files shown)


andreas-bok-sociomantic:
@leandro-lucarella-sociomantic , @ben-palmer-sociomantic 
Initial commit for adding deb packaging 

TODO's 
- config for ci services integration like `jenkins` or `travis`, cf. #11 
- currently the naming of the deb package is based on the status 
  of the git branch. Should this be kept or should we adopt a different 
  naming schema

leandro-lucarella-sociomantic:
Looks like a good first step to have a basic package. Hopefully eventually we'll get a DD interested in doing a proper package for this :wink: 

> config for ci services integration like jenkins or travis, cf. #11

Can travis archive build artefacts? Otherwise we'll have to leave this one for later.

> currently the naming of the deb package is based on the status of the git branch. Should this be kept or should we adopt a different naming schema

You mean the version, right? I think it's OK. Before a proper release we just have to do a proper annotated tag to get a clean version.

andreas-bok-sociomantic:
Yes, git branch status prob not the right term, but a name like this is what I meant. Dirty as the name says 

`libdrizzle-redux_5.1.4+6~dirty.20160108130828~ga987348-trusty_all.deb`

leandro-lucarella-sociomantic:
It's dirty because your repo is :wink: 

If you get a clean `git status` in a annotated tagged commit, you'll get a clean version in the package.

andreas-bok-sociomantic:
Hehe, ok. So we should make a tag on the next release. 

Given the guideline in `configure.ac` what would be a reasonable version number

**configure.ac**

```
...
#shared library versioning
LIBDRIZZLE_LIBRARY_VERSION=9:0:0
#                         | | |
#                  +------+ | +---+
#                  |        |     |
#                 current:revision:age
#                  |        |     |
#                  |        |     +- increment if interfaces have been added
#                  |        |        set to zero if interfaces have been removed or changed
#                  |        +- increment if source code has changed
#                  |           set to zero if current is incremented
#                  +- increment if interfaces have been added, removed or changed
...
```

ben-palmer-sociomantic:
> Given the guideline in configure.ac what would be a reasonable version number

It seems to me then that we should release with our change as `v5.2.4` as the source code has changed with the adding back of the signal functions.

ben-palmer-sociomantic:
I don't know heaps about this area but this LGTM. Once this is rebased we can do a new release and we should get a clean build of a deb package to add to our repos.

ben-palmer-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was a1bad129f3c080ebabb32c1117b6d42b6daf87be, new (rebased) HEAD is 67f3202e0536a8277acc9a27123cd838dc9aaa12

leandro-lucarella-sociomantic:
Beware, one thing in here is the source code / project version, and another is the shared object version. I have the feeling the quoted code refers to the later, which is the one used by the linker (added via `-soname`), and in the file name (like `libdrizzle.so.9.0.0`). In that sense I think we have to update the version in some other way.

More info: https://en.wikipedia.org/wiki/Soname

andreas-bok-sociomantic:
Yes, but would it be an idea to to align the the `so` name version number and the `release` version. Would make it less confusing when referring to the library.

ben-palmer-sociomantic:
> Beware, one thing in here is the source code / project version, and another is the shared object version.

Ah right I missed that. So in that case we would be going to:
`LIBDRIZZLE_LIBRARY_VERSION=9:1:0`
and release:
`v5.1.5`

or is it better to keep the values incremented synched (i.e. release `v5.2.4`)?

leandro-lucarella-sociomantic:
> Yes, but would it be an idea to to align the the so name version number and the release version. Would make it less confusing when referring to the library.

The thing is sometimes they get out of sync I think, because maybe ABI changes are introduced, but the API stays the same (I think). At this point I'm not sure syncing the versions is an option unless we bump the "source version" to 9.x, because decreasing the soname version would be too confusing and potentially could break stuff for existing users.

@LinuxJedi @BrianAker, maybe you can give us a hand here, any advice on how this should be handled?

LinuxJedi:
In general I was trying to keep the so versioning in-line with the rules in that configure.ac whereas the release versioning was following http://semver.org/

You are right though, semantic version is for API and so version is for ABI.

There is a lot of history as to why we are at the version numbers we are at, not a lot of it I remember right now. My best guess is we are at 9.x.x for the so because Drizzle was at v7.x before libdrizzle was forked away from it (and I forked again for redux). I think the jump to semantic version 5.x.x was quite arbitrary. There was already a libdrizzle 1.0 and a lesser-known 2.0. So I suspect I jumped straight to 5.0 because it sounded good :)

I think you would probably find if you tried to align these they would very quickly pull apart from each other. With MySQL version 5.6 the so version is 18 for example.

leandro-lucarella-sociomantic:
Thanks @LinuxJedi 

I'd say we should keep bumping the current versions as needed. I don't think the ABI vs API version is something we can get around anyway, so it will be easier and less confusing for everybody if we keep building on top of the current version numbers.

PR #16, opened 2016-01-12
Update AC_INIT in configure.ac

andreas-bok-sociomantic:
Change parameters git AC_INIT to point
to the repository at github

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,7 +6,7 @@
 # Use and distribution licensed under the BSD license.  See
 # the COPYING file in this directory for full text.
 
-AC_INIT([libdrizzle],[5.1.4],[http://bugs.launchpad.net/drizzle],[libdrizzle],[http://drizzle.org/])
+AC_INIT([libdrizzle],[5.1.4],[https://github.com/sociomantic/libdrizzle-redux/issues],[libdrizzle],[https://github.com/sociomantic/libdrizzle-redux])
 
 AC_CONFIG_AUX_DIR([build-aux])
 
--- a/tests/unit/version.c
+++ b/tests/unit/version.c
@@ -48,7 +48,8 @@ int main(void)
   ASSERT_TRUE(strtol(LIBDRIZZLE_VERSION_STRING, (char **) NULL, 10));
   ASSERT_TRUE(LIBDRIZZLE_VERSION_HEX > 0);
   ASSERT_TRUE(drizzle_bugreport());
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -6,7 +6,7 @@
 # Use and distribution licensed under the BSD license.  See
 # the COPYING file in this directory for full text.
 
-AC_INIT([libdrizzle],[5.1.4],[http://bugs.launchpad.net/drizzle],[libdrizzle],[http://drizzle.org/])
```
ben-palmer-sociomantic, line 9: So will this line need to change for every release as it currently has `[5.1.4]` in it?

andreas-bok-sociomantic, line 9: Hmm, yes that the question. Probably it could be a variable defined using a git tag or some other dynamically updated version number...

ben-palmer-sociomantic, line 9: Well if its possible to get it from the git tag that would be great but if not we can make changing it part of the release process. The problem is it would be very easy to forget.

leandro-lucarella-sociomantic, line 9: +1

andreas-bok-sociomantic, line 9: > Well if its possible to get it from the git tag that would be great but if not we can make changing it part of the release process. The problem is it would be very easy to forget.

I will look into that.

andreas-bok-sociomantic, line 9: > I will look into that.

Trying to define **version** dynamically in `AC_INIT` without success. 
Any idea how to do that?

leandro-lucarella-sociomantic, line 9: There is always the extremely ugly option of generating `configure.ac` but I guess it would be too meta, not sure it worth the trouble.

andreas-bok-sociomantic, line 9: AFAICS it is simply not possible to define version number dynamically, cf. section for **AM_INIT_AUTOMAKE** https://www.gnu.org/software/automake/manual/html_node/Public-Macros.html

ben-palmer-sociomantic, line 9: Right, in that case I will merge this and if we find a way to set this dynamically we can add it later.

ben-palmer-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was 4a87d27a04bc37aa8b3fa95480d9a6d376e89d1b, new (rebased) HEAD is 5cb8bb4e4dfdcc0fd2c804b0de09c8d81c2baff4

PR #20, opened 2016-02-09
Fix install path for deb packaging

andreas-bok-sociomantic:
When running `dpkg -i` on the generated deb package, the libraries are installed 
to `/install/usr/local/lib`. 
Install path should be `/usr/local/lib`

### DIFF
--- a/deb/build
+++ b/deb/build
@@ -38,5 +38,5 @@ fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
 	--depends python2.7 \
 	--depends "git (>=1.7.7)" \
 	--deb-changelog "$changelog" \
-	install/usr
-
+	--deb-no-default-config-files \
+	install/=/
--- a/deb/build
+++ b/deb/build
@@ -35,8 +35,9 @@ fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
 	--vendor 'Sociomantic Labs GmbH' \
 	--license 'Simplified BSD License' \
 	--category shared-lib \
-	--depends python2.7 \
-	--depends "git (>=1.7.7)" \
+	--depends autoconf \
+	--depends zlib1g-dev \
+	--depends libtool \
 	--deb-changelog "$changelog" \
 	--deb-no-default-config-files \
 	install/=/
--- a/deb/include.am
+++ b/deb/include.am
@@ -4,19 +4,18 @@
 # 
 # Makefile for building deb package 
 
+.PHONY: clean-deb deb install-deb release-deb
 
-.PHONY: deb
-deb: 
+deb: clean-deb
+	./configure --prefix=/usr
 	$(MAKE) DESTDIR=${abs_builddir}/deb/install install 
-	deb/build
+	deb/build deb
 
-.PHONY: install-deb
 install-deb:
 	@echo "This target is empty at the moment." 
 	@echo "Should be something like:" 
 	@echo "    sudo dpkg -i libdrizzle-redux.[version].d"
 
-.PHONY: release-deb
 release-deb:
 	@read -p "Enter version (previous: $$(git describe --abbrev=0)): " version; \
 	test -z $$version && exit 1; \
@@ -29,5 +28,4 @@ release-deb:
 	git tag -a -m "$$msg" $$version
 
 clean-deb:
-	-rm -rf deb/*.deb deb/install 
-
+	-rm -rf deb/*.deb deb/install


ben-palmer-sociomantic:
Apart from my small comment. LGTM.

deb/build
```diff
@@ -28,6 +28,7 @@ trap "rm -f '$changelog'; exit 1" INT TERM QUIT
 pkgname=libdrizzle-redux
 genchangelog "$pkgname" "$pkgversion" "$pkgmaint" > "$changelog"
 fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
+	--prefix=/ \
```
leandro-lucarella-sociomantic: Are you sure this shouldn't be `/usr`?

leandro-lucarella-sociomantic: Oh, this is for `fpm`, not `./configure`...

leandro-lucarella-sociomantic:
Usually what you should do for packaging is:

```
./configure --prefix=/usr
make DESTDIR=${abs_builddir}/install install
fpm --.... ${abs_builddir}/install=/
```

(untested, but I hope you get the intention)

andreas-bok-sociomantic:
> (untested, but I hope you get the intention)

Get the gist

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic 
Currently the symbolic linking of `libdrizzle.so -> libdrizzle.so.9.0.0` overwrites an existing installation of `libdrizzle`. Is there some way of controlling the symbolic linking of the libraries during install or make step.   

Currently this is the linking done in `/usr/local/lib` after installation

```
libdrizzle.so -> libdrizzle.so.9.0.0*
libdrizzle.so.9 -> libdrizzle.so.9.0.0*
libdrizzle.so.9.0.0*
```

leandro-lucarella-sociomantic:
Maybe we should name the library libdrizzle-redux instead? Is not exactly the same as the old libdrizzle, right?

andreas-bok-sociomantic:
> Maybe we should name the library libdrizzle-redux instead? Is not exactly the same as the old libdrizzle, right?

Yes probably better. Would resolve the issue of library name clashing

leandro-lucarella-sociomantic:
This way it would be possible/simpler also to have both installed in parallel (either for testing or if some project is using and another the other one).

mathias-baumann-sociomantic:
the `-dev` version of a library usually installs the symlink thus conflicting with another version of a `-dev` package of the same library. However, if we use libdrizzle-redux as a name for the actual files all of this should be fine

ben-palmer-sociomantic:
Ok, sounds like changing the name is the easiest option. @andreas-bok-sociomantic what is involved in changing this? Is it just a build target?

andreas-bok-sociomantic:
>  @andreas-bok-sociomantic what is involved in changing this? Is it just a build target?

I am looking at it now. Apparently doesn't use the library and linker name, specified in `configure.ac` when creating the shared library...

andreas-bok-sociomantic:
Created PR for renaming package to `libdrizzle-redux` #22

andreas-bok-sociomantic:
Minor change where install prefix was changed to `/usr`.

Any further comments? @mathias-baumann-sociomantic , @leandro-lucarella-sociomantic

deb/build
leandro-lucarella-sociomantic: I still don't see why this is needed.

andreas-bok-sociomantic:
> The commit message looks weird, it is really as intended?

Markdown syntax in commit messages...will strip that off. 

> I would still put all the installed files in a subdirectory, not directly in the deb/ directory (like deb/install) and then use install/=/ in fpm.

I will fix that tomorrow.

deb/build
andreas-bok-sociomantic: Found it isn't. Thought the default install path would be `/usr/local`

andreas-bok-sociomantic:
Addressed comments from @leandro-lucarella-sociomantic

deb/build
leandro-lucarella-sociomantic: fpm is quite silly, it will copy stuff wherever you tell it to copy them, I think that option is for other types of packages.

deb/build
```diff
@@ -35,8 +35,9 @@ fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
...
-	--depends python2.7 \
-	--depends "git (>=1.7.7)" \
+	--depends autoconf \
+	--depends zlib1g-dev \
+	--depends libtool \
 	--deb-changelog "$changelog" \
-	install/usr
-
+	--deb-no-default-config-files \
+	usr/lib usr/include
```
leandro-lucarella-sociomantic: Why not just specifying all the files? What if some new files are added in the future?

leandro-lucarella-sociomantic: I would simply use `install/=/` as I suggested before (removing the `-C install` you added above).

andreas-bok-sociomantic:
> I would simply use install/=/ as I suggested before (removing the -C install you added above).

Got a bit confused since I thought you meant that `install/=/` should be an option for `fpm` and not an argument

leandro-lucarella-sociomantic:
> >    I would simply use install/=/ as I suggested before (removing the -C install you added above).
> 
> Got a bit confused since I thought you meant that install/=/ should be an option for fpm and not an argument 

fpm uses a syntax similar to rsync to "copy" files inside the package: https://github.com/jordansissel/fpm/wiki/Source:-dir#mapping-requires-fpm--0440

andreas-bok-sociomantic:
> fpm uses a syntax similar to rsync to "copy" files inside the package:

Ah I see. 

Applied the changes wrt. `install/=/`

leandro-lucarella-sociomantic:
OK, if it's tested and working, LGTM :)

andreas-bok-sociomantic:
It has been tested on a docker trusty image.

andreas-bok-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was 70fe33632c92bd39db11dd65a890446b26bd806e, new (rebased) HEAD is a5b8a62558882c8335af85e98da0889597e35951

Issue #29, opened 2016-04-06
Use fpm to build RPM package


andreas-bok-sociomantic:
Build rpm package using fpm to simplify the build script.

andreas-bok-sociomantic:
WIP PR: #30

PR #30, opened 2016-04-06
Build RPM package with fpm

andreas-bok-sociomantic:
- modify rpm build script to use fpm
- delete obsolete files
- add newly rpm package files to .gitignore

addresses #29


ben-palmer-sociomantic:
Is it possible to have one copy of the build file for both the rpm and deb builds?

andreas-bok-sociomantic:
> Is it possible to have one copy of the build file for both the rpm and deb builds?

I will check with @leandro-lucarella-sociomantic on what would be the normal approach for building distributions.

Currently the build files are almost identical, in which case the some dist specific argument for fpm could be passed to `build` 

Also the original `rpm` build target was creating packages for multiple architectures, [`rpm/include.am`](https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/master/rpm/include.am)

leandro-lucarella-sociomantic:
I'm not sure really what would be the best approach for this. I consider including packaging on the "upstream" repo only a best-effort, as every distribution has their own conventions.

andreas-bok-sociomantic:
Blocked until more best practice for rpm packaging has been further investigated.

PR #42, opened 2016-08-09
Include memory optimizations from launchpad repo

andreas-bok-sociomantic:
This PR adds changes from https://code.launchpad.net/~linuxjedi/libdrizzle/5.1-optimize 

The changes was merged after the **v5.1.4** release which was used as base for this fork of `libdrizzle-redux`

### DIFF
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -120,6 +120,7 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
   uint64_t offset= 0;
   size_t size= 0;
   uint64_t wire_size;
+  uint16_t current_field;
 
   drizzle_return_t unused_ret;
   if (ret_ptr == NULL)
@@ -156,8 +157,26 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
 
   if (result->field_buffer == NULL)
   {
-    result->field_buffer= new (std::nothrow) char[(*total) + 1];
-    if (result->field_buffer == NULL)
+    result->field_buffer= new (std::nothrow) char*[result->column_count]();
+    result->field_buffer_sizes= new (std::nothrow) size_t[result->column_count]();
+  }
+
+  /* If we haven't got the whole field then current field hasn't been
+   * incremented yet */
+  if ((result->field_offset + result->field_size) != result->field_total)
+  {
+    current_field= result->field_current;
+  }
+  else
+  {
+    current_field= result->field_current-1;
+  }
+
+  if (result->field_buffer_sizes[current_field] < (*total) + 1)
+  {
+    result->field_buffer[current_field]= (drizzle_field_t) realloc(result->field_buffer[current_field], (*total) + 1);
+    result->field_buffer_sizes[current_field]= (*total) + 1;
+    if (result->field_buffer[current_field] == NULL)
     {
       drizzle_set_error(result->con, __func__, "Failed to allocate.");
       *ret_ptr= DRIZZLE_RETURN_MEMORY;
@@ -165,7 +184,7 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
     }
   }
 
-  memcpy(result->field_buffer + offset, field, size);
+  memcpy(result->field_buffer[current_field] + offset, field, size);
 
   while ((offset + size) != (*total))
   {
@@ -175,12 +194,19 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
       return NULL;
     }
     assert(wire_size == (uint64_t)*total);
+    if ((result->field_offset + result->field_size) != result->field_total)
+    {
+      current_field= result->field_current;
+    }
+    else
+    {
+      current_field= result->field_current-1;
+    }
 
-    memcpy(result->field_buffer + offset, field, size);
+    memcpy(result->field_buffer[current_field] + offset, field, size);
   }
 
-  field= result->field_buffer;
-  result->field_buffer= NULL;
+  field= result->field_buffer[current_field];
   field[*total]= 0;
 
   return field;
@@ -190,7 +216,7 @@ void drizzle_field_free(drizzle_field_t field)
 {
   if (field)
   {
-    delete[] field;
+    free(field);
   }
 }
 
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -172,6 +172,26 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
     current_field= result->field_current-1;
   }
 
+  /*
+    The code below pre-allocates each field in the buffer to a statically defined
+    size to reduce the number of mallocs at the cost of a memory overhead
+
+    Experimental results did confirm that the number of allocations is halved
+    but with the caveat that the number of bytes allocated is doubled.
+
+    Given that the improvement from lazy allocation is already five orders of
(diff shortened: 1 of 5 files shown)


andreas-bok-sociomantic:
Tested and ready to merge.

libdrizzle/field.cc
```diff
@@ -172,6 +172,12 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
     current_field= result->field_current-1;
   }
 
+  if (result->field_buffer_sizes[current_field] == 0)
+  {
+    result->field_buffer[current_field]= (drizzle_field_t) malloc(64*1024);
+    result->field_buffer_sizes[current_field]= 64*1024;
```
don-clugston-sociomantic: Allocating 64KB will definitely solve the realloc problem but it will use so much ram it will probably cause the machine to go into swap. 
I'm a bit disturbed by this. If the fields was only set once, and then reallocated every time it exceeded the size, it would only reallocate a few times at startup, and then it would never happen again, so I doubt you would notice.
This sounds as though it is being allocated very often.

andreas-bok-sociomantic:
The datatypes can be selected like this:

``` sql
SELECT
  DATA_TYPE 
FROM
  INFORMATION_SCHEMA.COLUMNS 
WHERE
  TABLE_SCHEMA = 'some.schema'
AND
  TABLE_NAME = 'some.table';
```

andreas-bok-sociomantic:
valgrind shows significant improvements for the memory optimized version compared to the current implementation 

The query result set was 71800 rows with an average row-size of 37789 bytes. 

The row with **dev (no prealloc)** denotes the changes in c526631c01890a5643891aeb080ef46ff293f2ca

``` cpp
  if (result->field_buffer_sizes[current_field] == 0)
  {
    result->field_buffer[current_field]= (drizzle_field_t) malloc(64*1024);
    result->field_buffer_sizes[current_field]= 64*1024;
  }


```

**function: drizzle_row_buffer**

|  | #allocs | #frees | bytes allocated |
| --- | --- | --- | --- |
| dev | 143,641 | 143,641 | 37,692,368 |
| dev (no prealloc) | 143,693 | 143,693 | 36,038,968 |
| master | 2,193,254 | 2,193,254 | 2,694,315,601 |

**function: drizzle_field_buffer**

|  | #allocs | #frees | bytes allocated |
| --- | --- | --- | --- |
| dev | 41 | 41 | 3,228,368 |
| dev (no prealloc) | 93 | 93 | 1,574,968 |
| master | 2,049,654 | 2,049,654 | 2,659,851,601 |

andreas-bok-sociomantic:
With the testing results, the only question whether we would like to keep the preallocation code, c526631c01890a5643891aeb080ef46ff293f2ca

don-clugston-sociomantic:
Reduction in allocations from 2 million to 93 looks pretty darn good to me.
But halving the number of allocations in exchange for doubling the memory consumption does not look like a win. I think the preallocation code should be removed. If it turns out that memory allocation is still a problem, something a bit more clever should be used.

andreas-bok-sociomantic:
> If it turns out that memory allocation is still a problem, something a bit more clever should be used. 

I will comment it out and make a note in the header file and commit message

andreas-bok-sociomantic:
Given the experimental results, the code has been updated with further discussion of the field buffering

``` diff
diff --git a/libdrizzle-5.1/field_client.h b/libdrizzle-5.1/field_client.h
index 84e2c57..fc9c69b 100644
--- a/libdrizzle-5.1/field_client.h
+++ b/libdrizzle-5.1/field_client.h
@@ -83,6 +83,14 @@ drizzle_field_t drizzle_field_read(drizzle_result_st *result, uint64_t *offset,
  * The return value is a newly allocated buffer and must be freed using
  * drizzle_field_free when no longer needed.
  *
+ * The buffering is done by keeping an array of char* pointers, each representing
+ * a field from a result row. Memory is (re)allocated lazily as required by the
+ * size of the field to buffer. Moreover the allocation is conservative and
+ * allocates 'new-size + 1' bytes.
+ *
+ * For further details please refer to the inline documentation in
+ * libdrizzle/field.cc
+ *
  * @param[in]  result  A result object
  * @param[in]  total   The total size of the field, to be written to by the function
  * @param[out] ret_ptr A pointer to a drizzle_return_t to store the return status into
diff --git a/libdrizzle/field.cc b/libdrizzle/field.cc
index 998c422..40083f3 100644
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -172,11 +172,25 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
     current_field= result->field_current-1;
   }

-  if (result->field_buffer_sizes[current_field] == 0)
-  {
-    result->field_buffer[current_field]= (drizzle_field_t) malloc(64*1024);
-    result->field_buffer_sizes[current_field]= 64*1024;
-  }
+  /*
+    The code below pre-allocates each field in the buffer to a statically defined
+    size to reduce the number of mallocs at the cost of a memory overhead
+
+    Experimental results did confirm that the number of allocations is halved
+    but with the caveat that the number of bytes allocated is doubled.
+
+    Given that the improvement from lazy allocation is already five orders of
+    magnitude, having minimal memory overhead is (in our opinion) more favorable
+
+    The code is kept for reference and in the case that other would like to
+    pursue this optimization further
+
+    if (result->field_buffer_sizes[current_field] == 0)
+    {
+      result->field_buffer[current_field]= (drizzle_field_t) malloc(64*1024);
+      result->field_buffer_sizes[current_field]= 64*1024;
+    }
+  */

   if (result->field_buffer_sizes[current_field] < (*total) + 1)
   {
```

libdrizzle/field.cc
```diff
@@ -156,16 +157,54 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
...
+    The code below pre-allocates each field in the buffer to a statically defined
+    size to reduce the number of mallocs at the cost of a memory overhead
+
+    Experimental results did confirm that the number of allocations is halved
+    but with the caveat that the number of bytes allocated is doubled.
+
+    Given that the improvement from lazy allocation is already five orders of
+    magnitude, having minimal memory overhead is (in our opinion) more favorable
+
+    The code is kept for reference and in the case that other would like to
```
don-clugston-sociomantic: s/other/others/

don-clugston-sociomantic:
Apart from the typo in the comment, LGTM

andreas-bok-sociomantic:
> Apart from the typo in the comment, LGTM

Mistakenly squashed two of the commits. Reverted that but otherwise didn't change anything.

andreas-bok-sociomantic:
Updated based on comments from @don-clugston-sociomantic

``` diff
diff --git a/libdrizzle/field.cc b/libdrizzle/field.cc
index 40083f3..fa24bf1 100644
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -182,7 +182,7 @@ drizzle_field_t drizzle_field_buffer(drizzle_result_st *result, size_t *total,
-    The code is kept for reference and in the case that other would like to
+    The code is kept for reference and in the case that others would like to
```

don-clugston-sociomantic:
LGTM

ben-palmer-sociomantic:
LGTM

ben-palmer-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was c827857942da342fc8e33c474a97a157851b02da, new (rebased) HEAD is 88f45b5e487a6b3e198f11824c9c92aed6fec2f9

Issue #90, opened 2016-12-15
v5.4.0 breaks binary compatibility


andreas-bok-sociomantic:
The **ABI** bump to `10.0.0` breaks the binary compatibility. 
However version `v5.4.0` is compatible with the previous minor version `v5.3.4`

PR #91, opened 2016-12-15
Fix binary compatibility issue

andreas-bok-sociomantic:
Avoid breaking breaking the binary compatibility
The ABI version is demoted from 10.0.0 to ~9.0.4~ 9.4.0

Addresses #90

### DIFF
--- a/deb/after_install.sh
+++ b/deb/after_install.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
+
+sudo ln -sf /usr/lib/libdrizzle-redux.so.10 /usr/lib/libdrizzle-redux.so.9
--- a/deb/build
+++ b/deb/build
@@ -59,6 +59,7 @@ fpm -s dir -t deb -n "$pkgname" -v "$pkgversion" \
     --depends libgcc1 \
     --deb-changelog "$changelog" \
     --deb-no-default-config-files \
+    --after-install after_install.sh \
     $so_libname.$libversion=/usr/lib/ \
     $so_libname.$lib_major=/usr/lib/
 


mathias-lang-sociomantic:
Shouldn't this target v5.4.x ?
Commit title has a typo.
You might also want to update the documentation (or we can do that later).
Why `.0.4` ?

andreas-bok-sociomantic:
> You might also want to update the documentation (or we can do that later). 

in `configure.ac`?

> Why .0.4 ?

Next in line given the previous minor release  `v5.3.4` which had ABI version `9.3.0`

mathias-lang-sociomantic:
> in configure.ac?

Yes.

> Next in line given the previous minor release v5.3.4 which had ABI version 9.3.0

The previous one was `9.0.0`. I think it's fine to match the minor release directly, but then you meant `9.4.0`, not `9.0.4` right ?

EDIT: NVM me, the issue is only in the P.R / commit message.

andreas-bok-sociomantic:
> The previous one was 9.0.0. I think it's fine to match the minor release directly, but then you meant 9.4.0, not 9.0.4 right ?

Well, latest release was on v5.3.x (`v5.3.4`), with version in configure.ac = `9.3.0`. This generated `libdrizzle-redux.so.9.0.3`. 
With version in configure.ac = `9.4.0` generates `libdrizzle-redux.so.9.0.4`. 
version in configure.ac = `9.0.1` generates `libdrizzle-redux.so.8.1.0`. Pretty random in my understanding.

andreas-bok-sociomantic:
Fixed typos

ben-palmer-sociomantic:
Is this the right thing to do here? It seems to me that according to the versioning scheme used by the library that binaries are _not_ compatible.

mathias-lang-sociomantic:
retarded libdrizzle

andreas-bok-sociomantic:
> Is this the right thing to do here? It seems to me that according to the versioning scheme used by the library that binaries are not compatible.

Yeah, it is a bit weird. The reason is less about standards and more about pragmatics. Apps linked with `so.9` would have to be recompiled with `so.10` and redeployed. 

> @mathias-lang-sociomantic why not make a symlink from `libdrizzle-redux.so.9` to `libdrizzle-redux 10.0.0`. 

Less invasive wrt. to v5.4.0. And introduce a more sane versioning with the next major release.

mathias-lang-sociomantic:
The symlink will work too for us. But we should really work that out in the future. Can it be done by the package ?

andreas-bok-sociomantic:
> Can it be done by the package ?

Yes. **fpm** provides an arg: 

`--after-install FILE A script to be run after package installation`

Will try it test that it works

andreas-bok-sociomantic:
Tested with binaries linking against so.9 and so.10 respectively. Works

mathias-lang-sociomantic (approved):

andreas-bok-sociomantic:
Ok feel free to merge

deb/after_install.sh
```diff
@@ -0,0 +1,7 @@
+## !/usr/bin/bash
```
gautam-kotian-sociomantic: That's not really a shebang line, is it?

andreas-bok-sociomantic:
fixed dubious shebang and changed `##` to `#` in `after_install.sh`

```diff
diff --git a/deb/after_install.sh b/deb/after_install.sh
index a4cffdd..e467911 100644
--- a/deb/after_install.sh
+++ b/deb/after_install.sh
@@ -1,7 +1,6 @@
-## !/usr/bin/bash
-##
-## Script which is run after install to ensure binary compatibility between
-## this and the previous version of libdrizzle-redux
-##
+# !/usr/bin/env bash
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
```

deb/after_install.sh
```diff
@@ -0,0 +1,6 @@
+# !/usr/bin/env bash
```
gautam-kotian-sociomantic: Does it work if there's a space between `#` & `!`?

gautam-kotian-sociomantic: Just confirmed. It doesn't.

EDIT: I meant that this script may work (it's named with a `.sh` extension after all). But in general, a shebang line containing a space between `#` & `!` fails to be interpreted correctly.

gautam-kotian-sociomantic: I'd suggest just using `#!/bin/bash`

deb/after_install.sh
```diff
@@ -0,0 +1,6 @@
...
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
+
+sudo ln -sf /usr/lib/libdrizzle-redux.so.10  /usr/lib/libdrizzle-redux.so.9
```
gautam-kotian-sociomantic: There are two spaces between the last two arguments.

andreas-bok-sociomantic:
Updated based on comments from @gautam-kotian-sociomantic 
```diff
diff --git a/deb/after_install.sh b/deb/after_install.sh
index e467911..0edda68 100644
--- a/deb/after_install.sh
+++ b/deb/after_install.sh
@@ -1,6 +1,6 @@
-# !/usr/bin/env bash
+#!/bin/bash
 #
 # Script which is run after install to ensure binary compatibility between
 # this and the previous version of libdrizzle-redux
 
-sudo ln -sf /usr/lib/libdrizzle-redux.so.10  /usr/lib/libdrizzle-redux.so.9
+sudo ln -sf /usr/lib/libdrizzle-redux.so.10 /usr/lib/libdrizzle-redux.so.9
```

gautam-kotian-sociomantic:
LGTM.

PR #94, opened 2016-12-16
Bump version to v5.4.1

andreas-bok-sociomantic:


andreas-bok-sociomantic:
Should be the last merge before the new release.

PR #97, opened 2017-01-16
Make drizzle_binlog_get_filename return end position

andreas-bok-sociomantic:
Modify the function so the end position of the file is returned

### DIFF
--- a/libdrizzle-5.1/binlog.h
+++ b/libdrizzle-5.1/binlog.h
@@ -203,23 +203,31 @@ DRIZZLE_API
 const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_type);
 
 /**
- * Get the name of a binlog-file
+ * Get the name and size of a binlog file in bytes
  *
  * Queries the database for a list of binlog files and copies the filename to
  * the passed buffer
- * If the file_index is invalid or no binlog files exist filename will contain an
- * empty string.
- * A valid file_index is in the range [-1 to (number of binlog files -1)]
+ *
+ * If the file_index is invalid or no binlog files exist filename will contain
+ * an empty string. A valid file_index is in the range
+ * [-1 to (number of binlog files -1)]
+ *
+ * The end_position will hold the size of the binlog file and can be used to
+ * start reading from the end of the binlog file when parsed to binlog_start
  *
  * The filename parameter is allocated by the function and needs to be
- * freed by the application when finished with.
+ * freed by the application when finished with. This is the case, regardless of
+ * the return status of the function.
  *
- * @param[in] con Drizzle structure previously initialized with drizzle_create().
- * @param[in,out] filename buffer to copy filename to
- * @param[in] file_index index of the binlog to retrieve.
+ * @param[in] con Drizzle structure previously initialized with drizzle_create()
+ * @param[in,out] filename Buffer to copy filename to
+ * @param[in,out] end_position Variable to save the size of the binlog file into
+ * @param[in] file_index Index of the binlog to retrieve.
+ * @return Standard drizzle return value
  */
 DRIZZLE_API
-void drizzle_binlog_get_filename(drizzle_st *con, char **filename, int file_index);
+drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
+                                             uint32_t *end_position, int file_index);
 
 #ifdef __cplusplus
 }
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -449,44 +449,54 @@ const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_typ
     }
 }
 
-void drizzle_binlog_get_filename(drizzle_st *con, char **filename, int file_index)
+drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
+                                             uint32_t *end_position, int file_index)
 {
+  *filename = (char*) malloc(1);
+  *filename[0] = '\0';
+  *end_position = 0;
   drizzle_return_t driz_ret;
+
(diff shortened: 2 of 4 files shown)


mathias-lang-sociomantic:
That's definitely not a point release. Actually it should even be a [major](https://community.kde.org/Policies/Binary_Compatibility_Examples#Change_the_return_type).

andreas-bok-sociomantic:
> That's definitely not a point release. Actually it should even be a major.

Add to the API yes, but doesn't break anything.

andreas-bok-sociomantic:
Would concede that it might should go into to a minor release.

andreas-bok-sociomantic:
- Changed the function signature so it returns a drizzle_return_t` value
- Added unittest of the changed function signature 
- Updated RST documentation 
- Rebased on latest v6.x.x

andreas-bok-sociomantic:
#103 should be merged first into v6.x.x

mathias-lang-sociomantic:
> #103 should be merged first into v6.x.x

Still targeting v5.x.x though ?...

andreas-bok-sociomantic:
> Still targeting v5.x.x though ?...

Fixed

andreas-bok-sociomantic:
More comments?

ben-palmer-sociomantic:
> More comments?

It is still blocked at the moment, so is it ready for review?

andreas-bok-sociomantic:
> It is still blocked at the moment, so is it ready for review?

Yes

libdrizzle-5.1/binlog.h
```diff
@@ -203,23 +203,32 @@ DRIZZLE_API
...
  * The filename parameter is allocated by the function and needs to be
- * freed by the application when finished with.
+ * freed by the application when finished with. This is the case, regardless of
+ * the return status of the function.
  *
- * @param[in] con Drizzle structure previously initialized with drizzle_create().
- * @param[in,out] filename buffer to copy filename to
- * @param[in] file_index index of the binlog to retrieve.
+ * @param[in] con Drizzle structure previously initialized with
+ * drizzle_create() or drizzle_clone().
```
ben-palmer-sociomantic: No more `drizzle_clone` right?

libdrizzle-5.1/binlog.h
```diff
@@ -203,23 +203,32 @@ DRIZZLE_API
...
+ * @param[in,out] filename Buffer to copy filename to
+ * @param[in,out] end_position Variable to save the size of the binlog file into
+ * @param[in] file_index: Index of the binlog to retrieve.
```
ben-palmer-sociomantic: Most of the parameters don't have `:` in them.

ben-palmer-sociomantic (commented):
I had a couple of comments on the documentation otherwise LGTM.

andreas-bok-sociomantic:
Addressed comments by @ben-palmer-sociomantic 

Before merging: 
We might consider another change so the client is required to provide a preallocated buffer to copy   the filename into. Would make mem leaks less likely.
@mathias-lang-sociomantic

ben-palmer-sociomantic:
> We might consider another change so the client is required to provide a preallocated buffer to copy the filename into. Would make mem leaks less likely.

Whats the way other methods in the library do it?

andreas-bok-sociomantic:
> Whats the way other methods in the library do it?

[`drizzle_escape_str`](https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v5.6.x/libdrizzle/query.cc#L60) doesn't take a preallocated buffer. 
That's the only other API method with similar behavior.

ben-palmer-sociomantic:
> drizzle_escape_str doesn't take a preallocated buffer. That's the only other API method with similar behavior.

In that case I am fine with keeping it as it is.

libdrizzle/binlog.cc
```diff
@@ -449,44 +449,52 @@ const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_typ
...
+    drizzle_log_info(con, __FILE_LINE_FUNC__, "Found binary log '%s' with filesize %d bytes",
+                     *filename, *end_position);
+  }
+  else
+  {
+    drizzle_log_info(con, __FILE_LINE_FUNC__, "No binary log files found");
   }
 
   drizzle_result_free(result);
+  return DRIZZLE_RETURN_OK;
```
ben-palmer-sociomantic, line 501: Is it possible that with `drizzle_result_free(result);` only being called here we could end up returning an error code without freeing the result?

ben-palmer-sociomantic:
Just one more comment then LGTM.

libdrizzle/binlog.cc
andreas-bok-sociomantic: Indeed, that would be the case. Will fix

andreas-bok-sociomantic:
Addressed comment from @ben-palmer-sociomantic about possible memory leak. 

```diff
diff --git a/libdrizzle/binlog.cc b/libdrizzle/binlog.cc
index 9d30a5a..8a27b92 100644
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -464,6 +464,7 @@ drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
   if (driz_ret != DRIZZLE_RETURN_OK)
   {
     drizzle_set_error(con, __FILE_LINE_FUNC__, "Query to retrieve binary logs failed");
+    drizzle_result_free(result);
     return driz_ret;
   }
 
@@ -473,6 +474,7 @@ drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
   if (file_index < -1 || file_index >= row_count)
   {
     drizzle_set_error(con, __FILE_LINE_FUNC__, "Invalid binlog file index: %d", file_index);
+    drizzle_result_free(result);
     return DRIZZLE_RETURN_INVALID_ARGUMENT;
   }
```

PR #98, opened 2017-02-06
Fix allocation bug in drizzle_binlog_get_filename

andreas-bok-sociomantic:
Allocate length of buffer to `|filename| + 1` to make
the string correctly '\0' terminated

Add test of `drizzle_binlog_get_filename` to the binlog unittest

### DIFF
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -493,9 +493,10 @@ void drizzle_binlog_get_filename(drizzle_st *con, char **filename, int file_inde
                             row_count - 1 : file_index;
 
     drizzle_row_t row = drizzle_row_index(result, row_index);
-    *filename =( char*)realloc(*filename, strlen(row[0]));
-    memcpy(*filename, row[0], strlen(row[0]));
-    *filename[strlen(row[0])] = '\0';
+    size_t len = strlen(row[0]);
+    *filename = (char*)realloc(*filename, len + 1);
+    memcpy(*filename, row[0], len);
+    (*filename)[len] = '\0';
   }
 
   drizzle_result_free(result);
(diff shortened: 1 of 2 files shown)


libdrizzle/binlog.cc
```diff
@@ -493,9 +493,9 @@ void drizzle_binlog_get_filename(drizzle_st *con, char **filename, int file_inde
                             row_count - 1 : file_index;
 
     drizzle_row_t row = drizzle_row_index(result, row_index);
-    *filename =( char*)realloc(*filename, strlen(row[0]));
+    *filename = (char*)realloc(*filename, strlen(row[0]) + 1);
     memcpy(*filename, row[0], strlen(row[0]));
-    *filename[strlen(row[0])] = '\0';
+    (*filename)[strlen(row[0])] = '\0';
```
ben-palmer-sociomantic: So previously this line would have been writing beyond the end of the filename array right?

andreas-bok-sociomantic: Yeah, which should have segfaulted....

ben-palmer-sociomantic: Is there a way to test the response is correct, perhaps in a unit test or something?

mathias-lang-sociomantic: `(*filename)[$-1] = '\0';` is way better

andreas-bok-sociomantic:
Updated based on PR comments

libdrizzle/binlog.cc
andreas-bok-sociomantic: Yeah. I have changed the binlog unittest so it uses the filename retrieved from a `drizzle_binlog_get_filename` call

ben-palmer-sociomantic: Can you do that in C?

mathias-lang-sociomantic: I forgot about this inferior language, but @andreas-bok-sociomantic reminded me.
The main point was not to have double call to `strlen`, which is addressed now.

ben-palmer-sociomantic:
LGTM

PR #104, opened 2017-02-14
Fix compile error -Werror=logical-op

andreas-bok-sociomantic:
Fixes compiler emitted warning 
```sh
libdrizzle/conn.cc: In function 'drizzle_return_t drizzle_state_write(drizzle_st*)':
libdrizzle/conn.cc:1411:11: error: logical 'or' of equal expressions [-Werror=logical-op]
           || errno == EWOULDBLOCK
```

### DIFF
--- a/libdrizzle/conn.cc
+++ b/libdrizzle/conn.cc
@@ -1403,11 +1403,15 @@ drizzle_return_t drizzle_state_write(drizzle_st *con)
     { }
     else if (write_size == -1)
     {
-      if (errno == EAGAIN
-#ifdef EWOULDBLOCK
-          || errno == EWOULDBLOCK
+
+// workaround for  Werror=logical-op compile time error
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69602#c16
+#if EAGAIN == EWOULDBLOCK
+  #define EAGAIN_OR_WOULDBLOCK(e) (e == EAGAIN)
+#else
+  #define EAGAIN_OR_WOULDBLOCK(e) (e == EAGAIN || e == EWOULDBLOCK)
 #endif
-          )
+      if ( EAGAIN_OR_WOULDBLOCK(errno) )
       {
         ret= drizzle_set_events(con, POLLOUT);
         if (ret != DRIZZLE_RETURN_OK)


libdrizzle/conn.cc
```diff
@@ -1403,11 +1403,14 @@ drizzle_return_t drizzle_state_write(drizzle_st *con)
     { }
     else if (write_size == -1)
     {
-      if (errno == EAGAIN
+
+// workaround for  Werror=logical-op compile time error
```
ben-palmer-sociomantic: Extra space between `for` and `Werror`

libdrizzle/conn.cc
```diff
@@ -1403,11 +1403,14 @@ drizzle_return_t drizzle_state_write(drizzle_st *con)
...
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69602#c7
 #ifdef EWOULDBLOCK
-          || errno == EWOULDBLOCK
+      if (0 || errno == EAGAIN || errno == EWOULDBLOCK)
```
ben-palmer-sociomantic: What does the `0` do here? Off the top of my head I can't see why it is needed.

ben-palmer-sociomantic (commented):
Finished review.

andreas-bok-sociomantic:
> What does the 0 do here? Off the top of my head I can't see why it is needed.

Indeed not very logical. Used the fix suggested in the thread concerning the GCC bug, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69602#c7

However looking more into the discussion a better (less ugly) solution is suggested 

```c++
  #if EAGAIN == EWOULDBLOCK 
    #define EAGAIN_OR_WOULDBLOCK(e) (e == EAGAIN)
  #else
     #define EAGAIN_OR_WOULDBLOCK(e) (e == EAGAIN || e == EWOULDBLOCK)
  #endif

  ...
  if (EAGAIN_OR_WOULDBLOCK (errno))
```

mathias-lang-sociomantic (approved):
New fix looks better. Compiler is still retarded :P

ben-palmer-sociomantic:
LGTM

PR #105, opened 2017-02-15
Adds c++11 compile flags if compiler supports it

andreas-bok-sociomantic:


ben-palmer-sociomantic:
I don't know much about this but it LGTM. Do you want to get a second opinion or should I merge?

andreas-bok-sociomantic:
I think it fine to merge this PR since the change is all part of the autoconf macro that is used.

PR #110, opened 2017-02-23
Use TCP keep alive to prevent connection freezing on slow connections

ben-palmer-sociomantic:
The mysql connection could indefinately freeze when in non-blocking mode
and on a bad internet connection. Set the TCP keep alive option to keep
the connection alive.

In addition the following other options are set:
Idle time in seconds after which to send keep alive probes - 3 seconds
Maximum number of keep alive probes before the connection is declared
dead and dropped - 3
Time in seconds between keep alive probes - 3

### DIFF
--- a/libdrizzle-5.1/conn.h
+++ b/libdrizzle-5.1/conn.h
@@ -141,6 +141,51 @@ DRIZZLE_API
 void drizzle_options_destroy(drizzle_options_st *options);
 
 /**
+ * Create a new socket options object with the specified values
+ *
+ * @param[in] _wait_timeout The timeout (in seconds) for setsockopt calls with
+ *  option values: SO_SNDTIMEO, SO_RCVTIMEO, SO_LINGER
+ * @param[in] _keepidle The time (in seconds) the connection needs to remain
+ *  idle before TCP starts sending keepalive probes
+ * @param[in] _keepcnt The maximum number of keepalive probes TCP should send
+ *  before dropping the connection.
+ * @param[in] _keepintvl The time (in seconds) between individual keepalive probes
+ * @return The new socket options object
+ */
+DRIZZLE_API
+drizzle_socket_options_st *drizzle_socket_options_create(int wait_timeout,
+                                                         int keepidle,
+                                                         int keepcnt,
+                                                         int keepintvl);
+
+/**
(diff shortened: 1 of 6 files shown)


ben-palmer-sociomantic:
Blocked until tested.

andreas-bok-sociomantic:
Could you change the base to **v5.4.x**?

ben-palmer-sociomantic:
> Could you change the base to v5.4.x?

Is it a bug fix though or a new feature? I was thinking a new feature.

andreas-bok-sociomantic:
> Is it a bug fix though or a new feature? I was thinking a new feature.

Ah true, v5.x.x is the right one then

andreas-bok-sociomantic:
Comparing with the source code in "classic" drizzle library, the KEEPALIVE option on the setsockopt is removed in the libdrizzle-redux fork. 
libdrizzle-redux  http://bazaar.launchpad.net/~drizzle-trunk/drizzle/7.2/view/head:/drizzled/plugin/listen_tcp.cc#L128

andreas-bok-sociomantic:
LGTM, just wondering if we should add the keep alive values as settable options.

gautam-kotian-sociomantic:
Typo in `indefinately` in commit message. LGTM otherwise.

ben-palmer-sociomantic:
> just wondering if we should add the keep alive values as settable options.

Yeah I think this is probably a good idea, so that will add four new options.

andreas-bok-sociomantic:
cool

andreas-bok-sociomantic:
Probably the best place to add it is on the drizzle_options_struct, https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v5.x.x/libdrizzle/structs.h#L154

With default values. Won't require separate setter methods or arguments to the constructor

andreas-bok-sociomantic:
Created a PR which allow the user to specify socket connection options, https://github.com/ben-palmer-sociomantic/libdrizzle-redux/pull/1

ben-palmer-sociomantic:
Updated with additional commits from @andreas-bok-sociomantic

ben-palmer-sociomantic:
Closing as this is replaced by https://github.com/sociomantic-tsunami/libdrizzle-redux/pull/114

PR #119, opened 2017-03-14
Bump version to v5.5.0

andreas-bok-sociomantic:
new ABI version is so.10.1.0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,7 +6,7 @@
 # Use and distribution licensed under the BSD license.  See
 # the COPYING file in this directory for full text.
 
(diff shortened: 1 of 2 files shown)


ben-palmer-sociomantic:
Should the commit message say the new ABI version is `11:0:1`?

andreas-bok-sociomantic:
> Should the commit message say the new ABI version is 11:0:1

No. The generated version is `10.1.0`, just the idiosyncrasy of `libtool` which brings confusion.

ben-palmer-sociomantic:
> No. The generated version is 10.1.0, just the idiosyncrasy of libtool which brings confusion.

Ok, I had a feeling it might be something like this. LGTM then.

Issue #120, opened 2017-03-20
Check for error packets while doing handshake


mathias-baumann-sociomantic:
It is possible to receive an error packet when doing the handshake, however the code doesn't check for that.

The check needs to be added here:
https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/42d7a60183dd2eb6ec9033447353ef36882e480b/libdrizzle/handshake.cc#L101

There already is code doing such checks for packets arriving after the handshake, found here:
https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/c07059be4eb5027c593391848165141c4ef4a655/libdrizzle/result.cc#L467-L481

See also the definition of the error packet here:
https://dev.mysql.com/doc/internals/en/packet-ERR_Packet.html

And just for anecdotal information, I found & fixed exactly the same buck in mysql-native: https://github.com/mysql-d/mysql-native/issues/20

andreas-bok-sociomantic:
Fixed by #121

PR #121, opened 2017-03-21
Bugfix handshake error packet

andreas-bok-sociomantic:
Check for error packet while doing handshake

If an error packet is received instead of the
expected handshake packet the function returns
DRIZZLE_RETURN_ERROR_CODE

Addresses #120

### DIFF
--- a/libdrizzle/pack.cc
+++ b/libdrizzle/pack.cc
@@ -481,3 +481,28 @@ static drizzle_return_t _pack_scramble_hash(drizzle_st *con,
 
   return DRIZZLE_RETURN_OK;
 }
+
+bool drizzle_check_unpack_error(drizzle_st *con)
+{
+  // Check if the first byte is 0xFF, i.e. we received an error packet
+  if (con->buffer_ptr[0] != 255)
+    return false;
+
+  con->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
+  /* Byte 3 is always a '#' character, skip it. */
+  memcpy(con->sqlstate, con->buffer_ptr + 4, DRIZZLE_MAX_SQLSTATE_SIZE);
+  con->buffer_ptr+= 9;
+  con->buffer_size-= 9;
+  con->packet_size-= 9;
+
+  snprintf(con->last_error, DRIZZLE_MAX_ERROR_SIZE, "%.*s",
+           (int)con->packet_size-1, con->buffer_ptr);
+
+  drizzle_set_error(con, __func__, " %s", con->last_error);
+
+  con->buffer_ptr+= con->packet_size;
+  con->buffer_size-= con->packet_size;
+  con->packet_size= 0;
+
+  return true;
+}
--- a/libdrizzle/pack.h
+++ b/libdrizzle/pack.h
@@ -96,6 +96,17 @@ drizzle_return_t drizzle_unpack_string(drizzle_st *con, char *buffer,
 unsigned char *drizzle_pack_auth(drizzle_st *con, unsigned char *ptr,
                            drizzle_return_t *ret_ptr);
 
+/**
+ * Check if a MySQL ERR_Packet has been received, i.e. the first byte of the
+ * received packet is 0xFF.
+ * If an error packet was received the error_code and error message are set
+ *
+ * @param[in] con Drizzle structure previously initialized with
+ *  drizzle_create() or drizzle_clone().
+ * @return True if an error packet was received, false otherwise
+ */
+bool drizzle_check_unpack_error(drizzle_st *con);
+
 /** @} */
 
 #ifdef __cplusplus
--- a/libdrizzle/result.cc
+++ b/libdrizzle/result.cc
@@ -464,19 +464,11 @@ drizzle_return_t drizzle_state_result_read(drizzle_st *con)
     con->packet_size-= 5;
     ret= DRIZZLE_RETURN_OK;
   }
-  else if (con->buffer_ptr[0] == 255)
+  else if (drizzle_check_unpack_error(con))
   {
-    con->result->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
-    con->error_code= con->result->error_code;
-    /* Byte 3 is always a '#' character, skip it. */
-    memcpy(con->result->sqlstate, con->buffer_ptr + 4,
+    memcpy(con->result->sqlstate, con->sqlstate,
            DRIZZLE_MAX_SQLSTATE_SIZE);
     con->result->sqlstate[DRIZZLE_MAX_SQLSTATE_SIZE]= 0;
-    memcpy(con->sqlstate, con->result->sqlstate,
-           DRIZZLE_MAX_SQLSTATE_SIZE + 1);
-    con->buffer_ptr+= 9;
-    con->buffer_size-= 9;
-    con->packet_size-= 9;
     ret= DRIZZLE_RETURN_ERROR_CODE;
   }
   else
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -297,29 +297,11 @@ drizzle_return_t drizzle_state_binlog_read(drizzle_st *con)
     con->binlog->error_fn(DRIZZLE_RETURN_EOF, con, con->binlog->binlog_context);
     return DRIZZLE_RETURN_EOF;
   }
-  else if (con->buffer_ptr[0] == 255)
+  else if (drizzle_check_unpack_error(con))
   {
-    con->result->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
-    con->error_code= con->result->error_code;
-    /* Byte 3 is always a '#' character, skip it. */
-    memcpy(con->result->sqlstate, con->buffer_ptr + 4,
-           DRIZZLE_MAX_SQLSTATE_SIZE);
-    con->result->sqlstate[DRIZZLE_MAX_SQLSTATE_SIZE]= 0;
-    memcpy(con->sqlstate, con->result->sqlstate,
-           DRIZZLE_MAX_SQLSTATE_SIZE + 1);
-    con->buffer_ptr+= 9;
-    con->buffer_size-= 9;
-    con->packet_size-= 9;
-
-    snprintf(con->last_error, DRIZZLE_MAX_ERROR_SIZE, "%.*s",
-             (int)con->packet_size, con->buffer_ptr);
-    con->last_error[DRIZZLE_MAX_ERROR_SIZE-1]= 0;
     snprintf(con->result->info, DRIZZLE_MAX_INFO_SIZE, "%.*s",
-             (int)con->packet_size, con->buffer_ptr);
+             (int)con->packet_size, con->last_error);
     con->result->info[DRIZZLE_MAX_INFO_SIZE-1]= 0;
-    con->buffer_ptr+= con->packet_size;
-    con->buffer_size-= con->packet_size;
-    con->packet_size= 0;
 
     con->pop_state();
     con->binlog->error_fn(DRIZZLE_RETURN_ERROR_CODE, con, con->binlog->binlog_context);
--- a/libdrizzle/handshake.cc
+++ b/libdrizzle/handshake.cc
@@ -99,7 +99,11 @@ drizzle_return_t drizzle_state_handshake_server_read(drizzle_st *con)
     return DRIZZLE_RETURN_OK;
   }
 
-  if (con->packet_size < 46)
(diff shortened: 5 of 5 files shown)


andreas-bok-sociomantic:
fyi @mathias-baumann-sociomantic

libdrizzle/pack.cc
```diff
@@ -481,3 +481,24 @@ static drizzle_return_t _pack_scramble_hash(drizzle_st *con,
 
   return DRIZZLE_RETURN_OK;
 }
+
+void drizzle_unpack_error(drizzle_st *con)
```
mathias-baumann-sociomantic: Does it make sense to call this function from https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/c07059be4eb5027c593391848165141c4ef4a655/libdrizzle/result.cc#L467-L481 to avoid code duplication?

andreas-bok-sociomantic: Yes. But might have to change the logic a bit while the currently `drizzle_unpack_error` sets the `packet_size = 0`, which is not done in the error parsing in `result.cc`

andreas-bok-sociomantic: Hmm, `con->buffer_size >= con->packet_size` is guaranteed so shouldn't be a problem to do 
```c++
  con->buffer_ptr+= con->packet_size;
  con->buffer_size-= con->packet_size;
  con->packet_size= 0;
```

andreas-bok-sociomantic:
Updated based on suggestions from @mathias-baumann-sociomantic 

```diff 
diff --git a/libdrizzle/binlog.cc b/libdrizzle/binlog.cc
index fc4a2ea..8708b58 100644
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -299,27 +299,10 @@ drizzle_return_t drizzle_state_binlog_read(drizzle_st *con)
   }
   else if (con->buffer_ptr[0] == 255)
   {
-    con->result->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
-    con->error_code= con->result->error_code;
-    /* Byte 3 is always a '#' character, skip it. */
-    memcpy(con->result->sqlstate, con->buffer_ptr + 4,
-           DRIZZLE_MAX_SQLSTATE_SIZE);
-    con->result->sqlstate[DRIZZLE_MAX_SQLSTATE_SIZE]= 0;
-    memcpy(con->sqlstate, con->result->sqlstate,
-           DRIZZLE_MAX_SQLSTATE_SIZE + 1);
-    con->buffer_ptr+= 9;
-    con->buffer_size-= 9;
-    con->packet_size-= 9;
-
-    snprintf(con->last_error, DRIZZLE_MAX_ERROR_SIZE, "%.*s",
-             (int)con->packet_size, con->buffer_ptr);
-    con->last_error[DRIZZLE_MAX_ERROR_SIZE-1]= 0;
+    drizzle_unpack_error(con);
     snprintf(con->result->info, DRIZZLE_MAX_INFO_SIZE, "%.*s",
-             (int)con->packet_size, con->buffer_ptr);
+             (int)con->packet_size, con->last_error);
     con->result->info[DRIZZLE_MAX_INFO_SIZE-1]= 0;
-    con->buffer_ptr+= con->packet_size;
-    con->buffer_size-= con->packet_size;
-    con->packet_size= 0;
 
     con->pop_state();
     con->binlog->error_fn(DRIZZLE_RETURN_ERROR_CODE, con, con->binlog->binlog_context);
diff --git a/libdrizzle/result.cc b/libdrizzle/result.cc
index b039218..33ee8b0 100644
--- a/libdrizzle/result.cc
+++ b/libdrizzle/result.cc
@@ -466,17 +466,10 @@ drizzle_return_t drizzle_state_result_read(drizzle_st *con)
   }
   else if (con->buffer_ptr[0] == 255)
   {
-    con->result->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
-    con->error_code= con->result->error_code;
-    /* Byte 3 is always a '#' character, skip it. */
-    memcpy(con->result->sqlstate, con->buffer_ptr + 4,
+    drizzle_unpack_error(con);
+    memcpy(con->result->sqlstate, con->sqlstate,
            DRIZZLE_MAX_SQLSTATE_SIZE);
     con->result->sqlstate[DRIZZLE_MAX_SQLSTATE_SIZE]= 0;
-    memcpy(con->sqlstate, con->result->sqlstate,
-           DRIZZLE_MAX_SQLSTATE_SIZE + 1);
-    con->buffer_ptr+= 9;
-    con->buffer_size-= 9;
-    con->packet_size-= 9;
     ret= DRIZZLE_RETURN_ERROR_CODE;
   }
   else

```

libdrizzle/pack.cc
```diff
@@ -481,3 +481,24 @@ static drizzle_return_t _pack_scramble_hash(drizzle_st *con,
...
+{
+  assert(con->buffer_ptr[0] == 255);
```
mathias-lang-sociomantic: I'd say make it `drizzle_check_unpack_error` and return `false` if it's `!= 255`.
This way you loose the assert, and you can use it in `if`. Also less usage of the magic `255`.

libdrizzle/pack.h
```diff
@@ -96,6 +96,15 @@ drizzle_return_t drizzle_unpack_string(drizzle_st *con, char *buffer,
...
                            drizzle_return_t *ret_ptr);
 
+/**
+ * @brief           Unpack a MySQL ERR_Packet and set error_code and error
+ *                  message on the connection
+ *
+ * @param[in]       con         drizzle Drizzle structure previously initialized
+ *                              with drizzle_create() or drizzle_clone().
+ */
+void drizzle_unpack_error(drizzle_st *con);
```
mathias-lang-sociomantic: Does it means it's public ? Does it need to be ?

andreas-bok-sociomantic: `pack.h` is not a client include header

andreas-bok-sociomantic:
Something along these lines, @mathias-lang-sociomantic

mathias-lang-sociomantic (commented):
Not familiar with the internal, but the approach LGTM

andreas-bok-sociomantic:
> Not familiar with the internal, but the approach LGTM

OK, I will squash the commits

andreas-bok-sociomantic:
Squashed commits and updated func documentation 
```diff
diff --git a/libdrizzle/pack.h b/libdrizzle/pack.h
index 8ec58b4..fb520e5 100644
--- a/libdrizzle/pack.h
+++ b/libdrizzle/pack.h
@@ -97,11 +97,12 @@ unsigned char *drizzle_pack_auth(drizzle_st *con, unsigned char *ptr,
                            drizzle_return_t *ret_ptr);
 
 /**
- * @brief           Unpack a MySQL ERR_Packet and set error_code and error
- *                  message on the connection
+ * @brief      Unpack a MySQL ERR_Packet and set error_code and error
+ *             message on the connection
  *
- * @param[in]       con         drizzle Drizzle structure previously initialized
- *                              with drizzle_create() or drizzle_clone().
+ * @param[in]  con         drizzle Drizzle structure previously initialized
+ *                         with drizzle_create() or drizzle_clone().
+ * @return     true if the packet is a MySQL Err_packet, false otherwise
  */
 bool drizzle_check_unpack_error(drizzle_st *con);
```

libdrizzle/pack.h
```diff
@@ -96,6 +96,16 @@ drizzle_return_t drizzle_unpack_string(drizzle_st *con, char *buffer,
 unsigned char *drizzle_pack_auth(drizzle_st *con, unsigned char *ptr,
                            drizzle_return_t *ret_ptr);
 
+/**
+ * @brief      Unpack a MySQL ERR_Packet and set error_code and error
+ *             message on the connection
+ *
+ * @param[in]  con         drizzle Drizzle structure previously initialized
```
ben-palmer-sociomantic: I don't think the whitespace for the in param matches the normal format.

libdrizzle/pack.cc
```diff
@@ -481,3 +481,27 @@ static drizzle_return_t _pack_scramble_hash(drizzle_st *con,
 
   return DRIZZLE_RETURN_OK;
 }
+
+bool drizzle_check_unpack_error(drizzle_st *con)
+{
+  if (con->buffer_ptr[0] != 255)
```
ben-palmer-sociomantic: Do we have a constant for this somewhere?

libdrizzle/result.cc
```diff
@@ -464,19 +464,11 @@ drizzle_return_t drizzle_state_result_read(drizzle_st *con)
...
     ret= DRIZZLE_RETURN_OK;
   }
-  else if (con->buffer_ptr[0] == 255)
+  else if (drizzle_check_unpack_error(con))
   {
-    con->result->error_code= drizzle_get_byte2(con->buffer_ptr + 1);
-    con->error_code= con->result->error_code;
-    /* Byte 3 is always a '#' character, skip it. */
-    memcpy(con->result->sqlstate, con->buffer_ptr + 4,
+    memcpy(con->result->sqlstate, con->sqlstate,
```
ben-palmer-sociomantic, line 469: Why are we changing this `memcpy`? Previously `con->sqlstate` was copied to `con->sqlstate`.

ben-palmer-sociomantic (commented):
Finished my review.

libdrizzle/result.cc
andreas-bok-sociomantic: It is changed because in `drizzle_check_unpack_error` mutates `con->buffer_ptr, con->buffer_size, 
  con->packet_size;`. Thus we would have to do some extra pointer arithmetics to copy the sqlstate from the `con->buffer_ptr` after the call `drizzle_check_unpack_error`

ben-palmer-sociomantic: Ok, no problem, just wanted to check the change was deliberate.

libdrizzle/pack.cc
andreas-bok-sociomantic: Not ATM. The values `(0, 254, 255)` are used as flags in various places but the meaning differs with the context. **PACKET_FLAG_0xFF** could work but doesn't connote the relation to the Err_packet

andreas-bok-sociomantic: Opted to keep 255 as literal number, and add a comment about its meaning here

andreas-bok-sociomantic:
Addressed comments from @ben-palmer-sociomantic 

```diff
diff --git a/libdrizzle/pack.cc b/libdrizzle/pack.cc
index fec4dd0..cd5bf01 100644
--- a/libdrizzle/pack.cc
+++ b/libdrizzle/pack.cc
@@ -484,6 +484,7 @@ static drizzle_return_t _pack_scramble_hash(drizzle_st *con,
 
 bool drizzle_check_unpack_error(drizzle_st *con)
 {
+  // Check if the first byte is 0xFF, i.e. we received an error packet
   if (con->buffer_ptr[0] != 255)
     return false;
 
@@ -497,7 +498,7 @@ bool drizzle_check_unpack_error(drizzle_st *con)
   snprintf(con->last_error, DRIZZLE_MAX_ERROR_SIZE, "%.*s",
            (int)con->packet_size-1, con->buffer_ptr);
 
-  drizzle_log_error(con, "%s %s", __func__, con->last_error);
+  drizzle_set_error(con, __func__, " %s", con->last_error);
 
   con->buffer_ptr+= con->packet_size;
   con->buffer_size-= con->packet_size;
diff --git a/libdrizzle/pack.h b/libdrizzle/pack.h
index fb520e5..77904b9 100644
--- a/libdrizzle/pack.h
+++ b/libdrizzle/pack.h
@@ -97,12 +97,13 @@ unsigned char *drizzle_pack_auth(drizzle_st *con, unsigned char *ptr,
                            drizzle_return_t *ret_ptr);
 
 /**
- * @brief      Unpack a MySQL ERR_Packet and set error_code and error
- *             message on the connection
+ * Check if a MySQL ERR_Packet has been received, i.e. the first byte of the
+ * received packet is 0xFF.
+ * If an error packet was received the error_code and error message are set
  *
- * @param[in]  con         drizzle Drizzle structure previously initialized
- *                         with drizzle_create() or drizzle_clone().
- * @return     true if the packet is a MySQL Err_packet, false otherwise
+ * @param[in] con Drizzle structure previously initialized with
+ *  drizzle_create() or drizzle_clone().
+ * @return True if an error packet was received, false otherwise
  */
 bool drizzle_check_unpack_error(drizzle_st *con);
```

ben-palmer-sociomantic:
LGTM, anything else @mathias-lang-sociomantic ?

mathias-lang-sociomantic (approved):
LGTM

PR #124, opened 2017-03-24
Expand and fix string escaping capabilities

mathias-lang-sociomantic:
The mysql manual lists 11 characters which should be escaped,
including 2 which should only escaped for string using in pattern-matching context (e.g. LIKE clause).
However, `drizzle_escape_string` was only escaping 7 characters.
As changing `drizzle_escape_string` is not an option, a new `drizzle_escape_str` is introduced
which takes an extra parameter, used as a boolean, for specifying whether the string will be
used in a pattern-matching context.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,3 +15,16 @@ SSL connections.
 New Features
 ============
 
+* `drizzle_escape_str`
+
+    This new function matches the signature of `drizzle_escape_string`, except it takes an extra parameter
+    to determine whether to escape pattern-matching characters ('%' and '_') or not.
+    Calling `drizzle_escape_string` is equivalent to calling `drizzle_escape_str` with `false` as last parameter.
+
+
+Bugs fixed
+==========
+
+- `drizzle_escape_string`
+
+    This function now correctly escapes `\b` and `\t`
--- a/libdrizzle-5.1/query.h
+++ b/libdrizzle-5.1/query.h
@@ -78,6 +78,9 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  * Escape a string for an SQL query. The to parameter is allocated by the
  * function and needs to be freed by the application when finished with.
  *
+ * This function does not escape '%' or '_', if you need to, call
+ * drizzle_escape_str with `is_pattern` set to `true`.
+ *
  * @param[in] con a connection object
  * @param[in,out] to the destination string
  * @param[in] from the source string
@@ -87,6 +90,31 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  */
 DRIZZLE_API
 ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, const size_t from_size);
+
+/**
+ * Escape a string for an SQL query, optionally for pattern matching.
+ *
+ * This function escapes the following characters:
+ * '\0' (0x00), '\'' (0x27), '"' (0x22), '\b' (0x08), '\n' (0x0A),
+ * '\r' (0x0D), '\t' (0x09), '\Z' (0x26), '\\' (0x5C).
+ * In case `is_pattern` is set to `true`, '%' (0x25) and '_' (0x5F)
+ * will be escaped as well.
+ *
+ * The to parameter is allocated by the function and needs to be freed by
+ * the application when finished with.
+ *
+ * @param[in] con a connection object
+ * @param[in,out] to the destination string
+ * @param[in] from the source string
+ * @param[in] from_size the length of the source string
+ * @param[in] is_pattern whether to escape '%' and '_'.
+ *                       If set to `true`, they will be escaped,
+ *                       so the string can be used in a LIKE clause for example.
+ * @return the length of the ‘to’ string or -1 upon error due to empty
+ *         parameters or overflow
+ */
+DRIZZLE_API
+ssize_t drizzle_escape_str(drizzle_st *con, char **to, const char *from, const size_t from_size, bool is_pattern);
 
 /** @} */
 
--- a/tests/unit/escape.c
+++ b/tests/unit/escape.c
@@ -52,16 +52,33 @@ int main(int argc, char* argv[])
 
   // Test for data too long
   char *out;
-  uint64_t out_len = drizzle_escape_string(NULL, &out, in, strlen(in));
+  uint64_t out_len = drizzle_escape_str(NULL, &out, in, strlen(in), false);
   ASSERT_EQ_(17, out_len, "drizzle_escape_string(): %u != %u", 17,
              (unsigned int)(out_len));
 
   free(out);
 
-  out_len= drizzle_escape_string(NULL, &out, in, strlen(in));
+  out_len= drizzle_escape_str(NULL, &out, in, strlen(in), false);
   ASSERT_EQ_(out_len, 17, "Bad hex length output %u", (unsigned int)(out_len));
   ASSERT_EQ_(0, strcmp(out, "escape \\\"this\\\"\\n"), "Bad hex data output");
 
+  free(out);
+
+  const char *with_meta = "This will trip % LIKE _ up";
+  out_len = drizzle_escape_str(NULL, &out, with_meta, strlen(with_meta), false);
+  ASSERT_EQ_(out_len, 26, "Bad hex length output %u", (unsigned int)(out_len));
+  ASSERT_EQ_(0, strcmp(out, with_meta), "Bad hex data output");
+  free(out);
+
+  out_len = drizzle_escape_str(NULL, &out, with_meta, strlen(with_meta), true);
+  ASSERT_EQ_(out_len, 28, "Bad hex length output %u", (unsigned int)(out_len));
+  ASSERT_EQ_(0, strcmp(out, "This will trip \\% LIKE \\_ up"), "Bad hex data output");
+  free(out);
+
+  const char *tricky = "This\tis \\\\\b tricky";
+  out_len = drizzle_escape_str(NULL, &out, tricky, strlen(tricky), false);
+  ASSERT_EQ_(out_len, 22, "Bad hex length output %u", (unsigned int)(out_len));
+  ASSERT_EQ_(0, strcmp(out, "This\\tis \\\\\\\\\\b tricky"), "Bad hex data output");
   free(out);
 
   return EXIT_SUCCESS;
--- a/libdrizzle/query.cc
+++ b/libdrizzle/query.cc
@@ -57,7 +57,7 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
                                    (unsigned char *)query, size, size, ret_ptr);
 }
 
-ssize_t drizzle_escape_string(drizzle_st *con, char **destination, const char *from, const size_t from_size)
+ssize_t drizzle_escape_str(drizzle_st *con, char **destination, const char *from, const size_t from_size, bool is_pattern)
 {
   (void)con;
   const char *end;
@@ -92,6 +92,12 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **destination, const char *f
       case 0:
         newchar= '0';
         break;
+      case '\b':
+        newchar= 'b';
+        break;
+      case '\t':
+        newchar= 't';
+        break;
       case '\n':
         newchar= 'n';
         break;
@@ -110,6 +116,14 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **destination, const char *f
       case '"':
         newchar= '"';
         break;
+      case '%':
+          if (is_pattern)
+              newchar = '%';
+          break;
+      case '_':
+          if (is_pattern)
+              newchar = '_';
+          break;
       default:
         break;
       }
@@ -144,6 +158,11 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **destination, const char *f
   *to= 0;
 
   return to_size;
+}
+
+ssize_t drizzle_escape_string(drizzle_st *con, char **destination, const char *from, const size_t from_size)
+{
+    return drizzle_escape_str(con, destination, from, from_size, false);
 }
 
 bool drizzle_hex_string(char *to, const unsigned char *from, const size_t from_size)


mathias-lang-sociomantic:
Reference: https://dev.mysql.com/doc/refman/5.7/en/string-literals.html

andreas-bok-sociomantic:
LGTM, 
I will add it to milestone [v6.0.0](https://github.com/sociomantic-tsunami/libdrizzle-redux/milestone/11). 
And then push for that release

RELEASE_NOTES.md
```diff
@@ -15,3 +15,16 @@ SSL connections.
 New Features
 ============
 
+* `drizzle_escape_str`
+
+    This new function matches the signature of `drizzle_escape_string`, except it takes an extra parameter
+    to determine whether to escape pattern-matching characters ('%' and '_') or not.
+    Calling `drizzle_escape_string` is equivalent to calling `drizzle_escape_str` with `0` as last parameter.
```
ben-palmer-sociomantic: You have weirdly managed to put the text in italics here but it doesn't seem to affect the final result.

libdrizzle-5.1/query.h
```diff
@@ -78,6 +78,9 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  * Escape a string for an SQL query. The to parameter is allocated by the
  * function and needs to be freed by the application when finished with.
  *
+ * This function does not escape '%' or '_', if you need to, call
+ * drizzle_escape_str with search set to 'true'.
```
ben-palmer-sociomantic: Should this be "with is_pattern set to true"?

libdrizzle-5.1/query.h
```diff
@@ -88,6 +91,31 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
...
 ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, const size_t from_size);
 
+/**
+ * Escape a string for an SQL query, optionally for pattern matching.
+ *
+ * This function escapes the following characters:
+ * '\0' (0x00), '\'' (0x27), '"' (0x22), '\b' (0x08), '\n' (0x0A),
+ * '\r' (0x0D), '\t' (0x09), '\Z' (0x26), '\\' (0x5C).
+ * In case `is_pattern` is set to a non-zero, '%' (0x25) and '_' (0x5F)
+ * wil be escaped as well.
```
ben-palmer-sociomantic: Spelling of `wil`

ben-palmer-sociomantic (commented):
Just some minor comments on the documentation from me.

ben-palmer-sociomantic:
> As changing drizzle_escape_string is not an option

Why is this? As this is a major version can't we change the API?

mathias-lang-sociomantic:
> Why is this? As this is a major version can't we change the API?

It wasn't targeted as a major version, but a minor.

ben-palmer-sociomantic:
> It wasn't targeted as a major version, but a minor.

Ah my confusion is that the milestone is set as `v6.0.0` but the target is `v5.x.x`.

andreas-bok-sociomantic:
My bad, will add a minor release milestone, `v5.6.0`

ben-palmer-sociomantic:
> My bad, will add a minor release milestone, v5.6.0

Do we need that? If v6.0.0 should be released soon why not just add the breaking changes to that?

andreas-bok-sociomantic:
> Do we need that? If v6.0.0 should be released soon why not just add the breaking changes to that?

@mathias-lang-sociomantic

mathias-lang-sociomantic:
> Do we need that? If v6.0.0 should be released soon why not just add the breaking changes to that?

Because that means you'll stop developing on v5.x.x straight away. And that's inconvenient, for the same reason we keep developing on ocean for 6 months before switching to the new major.

ben-palmer-sociomantic:
> Because that means you'll stop developing on v5.x.x straight away. And that's inconvenient, for the same reason we keep developing on ocean for 6 months before switching to the new major.

But we need to do a v6.0.0 release anyway for the required binlog changes.

mathias-lang-sociomantic:
> But we need to do a v6.0.0 release anyway for the required binlog changes.

So that means we completely drop v5.x.x support as soon as v6.0.0 is out ?

andreas-bok-sociomantic:
> Yes, So that means we completely drop v5.x.x support as soon as v6.0.0 is out?

That would be preferable, since v6.0.0 changes the directory layout. Merging from v5.x into v6.x woulod be a pain in the ass

libdrizzle-5.1/query.h
```diff
@@ -88,6 +91,31 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
...
+ * @param[in] from the source string
+ * @param[in] from_size the length of the source string
+ * @param[in] is_pattern whether to escape '%' and '_'.
+ *                   If set to a non-zero value, they will be escaped,
+ *                   so the string can be used in a LIKE clause for example.
+ * @return the length of the ‘to’ string or -1 upon error due to empty
+ *         parameters or overflow
+ */
+DRIZZLE_API
+ssize_t drizzle_escape_str(drizzle_st *con, char **to, const char *from, const size_t from_size, int is_pattern);
```
andreas-bok-sociomantic: why `int is_pattern` in the `query.h` and `bool is_pattern` in `query.cc`?

andreas-bok-sociomantic (commented):
Would be nice to add test of the extended capabilities to the existing [escape](https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v5.5.1/tests/unit/escape.c) unittest

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/libdrizzle-5.1/query.h b/libdrizzle-5.1/query.h
index 66f521f..e79d39e 100644
--- a/libdrizzle-5.1/query.h
+++ b/libdrizzle-5.1/query.h
@@ -79,7 +79,7 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  * function and needs to be freed by the application when finished with.
  *
  * This function does not escape '%' or '_', if you need to, call
- * drizzle_escape_str with search set to 'true'.
+ * drizzle_escape_str with `is_pattern` set to a non-zero value.
  *
  * @param[in] con a connection object
  * @param[in,out] to the destination string
@@ -98,7 +98,7 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, cons
  * '\0' (0x00), '\'' (0x27), '"' (0x22), '\b' (0x08), '\n' (0x0A),
  * '\r' (0x0D), '\t' (0x09), '\Z' (0x26), '\\' (0x5C).
  * In case `is_pattern` is set to a non-zero, '%' (0x25) and '_' (0x5F)
- * wil be escaped as well.
+ * will be escaped as well.
  *
  * The to parameter is allocated by the function and needs to be freed by
  * the application when finished with.
diff --git a/libdrizzle/query.cc b/libdrizzle/query.cc
index 6346e84..f83ec0e 100644
--- a/libdrizzle/query.cc
+++ b/libdrizzle/query.cc
@@ -57,7 +57,7 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
                                    (unsigned char *)query, size, size, ret_ptr);
 }
 
-ssize_t drizzle_escape_str(drizzle_st *con, char **destination, const char *from, const size_t from_size, bool is_pattern)
+ssize_t drizzle_escape_str(drizzle_st *con, char **destination, const char *from, const size_t from_size, int is_pattern)
 {
   (void)con;
   const char *end;
```

> Would be nice to add test of the extended capabilities to the existing escape unittest

Deal.

mathias-lang-sociomantic:
Added tests

andreas-bok-sociomantic:
LGTM, with the rest of the comments addressed it should be good to go

libdrizzle-5.1/query.h
andreas-bok-sociomantic: Ok, both are int now. 
Why not bool?

andreas-bok-sociomantic (commented):
Just one small comment. OLGTM

libdrizzle-5.1/query.h
mathias-lang-sociomantic: Because there's no bool in C ?

andreas-bok-sociomantic: Hmm, `bool` was added in the C99 specification. Furthermore, other API functions take/return `bool`.

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/libdrizzle-5.1/query.h b/libdrizzle-5.1/query.h
index e79d39e..c2c9844 100644
--- a/libdrizzle-5.1/query.h
+++ b/libdrizzle-5.1/query.h
@@ -108,13 +108,13 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, cons
  * @param[in] from the source string
  * @param[in] from_size the length of the source string
  * @param[in] is_pattern whether to escape '%' and '_'.
- *                   If set to a non-zero value, they will be escaped,
+ *                   If set to `true`, they will be escaped,
  *                   so the string can be used in a LIKE clause for example.
  * @return the length of the ‘to’ string or -1 upon error due to empty
  *         parameters or overflow
  */
 DRIZZLE_API
-ssize_t drizzle_escape_str(drizzle_st *con, char **to, const char *from, const size_t from_size, int is_pattern);
+ssize_t drizzle_escape_str(drizzle_st *con, char **to, const char *from, const size_t from_size, bool is_pattern);
 
 /** @} */
 
diff --git a/libdrizzle/query.cc b/libdrizzle/query.cc
index f83ec0e..6346e84 100644
--- a/libdrizzle/query.cc
+++ b/libdrizzle/query.cc
@@ -57,7 +57,7 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
                                    (unsigned char *)query, size, size, ret_ptr);
 }
 
-ssize_t drizzle_escape_str(drizzle_st *con, char **destination, const char *from, const size_t from_size, int is_pattern)
+ssize_t drizzle_escape_str(drizzle_st *con, char **destination, const char *from, const size_t from_size, bool is_pattern)
 {
   (void)con;
   const char *end;
```

RELEASE_NOTES.md
ben-palmer-sociomantic: Should this still be "with `0` as last parameter."?

libdrizzle-5.1/query.h
```diff
@@ -88,6 +91,31 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
 DRIZZLE_API
 ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, const size_t from_size);
 
+/**
+ * Escape a string for an SQL query, optionally for pattern matching.
+ *
+ * This function escapes the following characters:
+ * '\0' (0x00), '\'' (0x27), '"' (0x22), '\b' (0x08), '\n' (0x0A),
+ * '\r' (0x0D), '\t' (0x09), '\Z' (0x26), '\\' (0x5C).
+ * In case `is_pattern` is set to a non-zero, '%' (0x25) and '_' (0x5F)
```
andreas-bok-sociomantic: `s/a non-zero/true/`

libdrizzle-5.1/query.h
```diff
@@ -78,6 +78,9 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  * Escape a string for an SQL query. The to parameter is allocated by the
  * function and needs to be freed by the application when finished with.
  *
+ * This function does not escape '%' or '_', if you need to, call
+ * drizzle_escape_str with `is_pattern` set to a non-zero value.
```
ben-palmer-sociomantic: Set to `true`?

libdrizzle-5.1/query.h
ben-palmer-sociomantic: "non-zero" should probably be "true" here as well

libdrizzle-5.1/query.h
andreas-bok-sociomantic: `s/a non-zero value/true/`

ben-palmer-sociomantic:
In the commit message:
"including 2 which should only escaped for string using in pattern-matching context (e.g. LIKE clause)."
would be better as:
"including 2 which should only be escaped for strings used in a pattern-matching context (e.g. LIKE clause)."

and "used as a boolean" can probably be removed.

mathias-lang-sociomantic:
Oh damn, I didn't realized I was repeating myself so much :P

Updated:
```diff
diff --git a/RELEASE_NOTES.md b/RELEASE_NOTES.md
index 709e754..753908f 100644
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -19,7 +19,7 @@ New Features
 
     This new function matches the signature of `drizzle_escape_string`, except it takes an extra parameter
     to determine whether to escape pattern-matching characters ('%' and '_') or not.
-    Calling `drizzle_escape_string` is equivalent to calling `drizzle_escape_str` with `0` as last parameter.
+    Calling `drizzle_escape_string` is equivalent to calling `drizzle_escape_str` with `false` as last parameter.
 
 
 Bugs fixed
diff --git a/libdrizzle-5.1/query.h b/libdrizzle-5.1/query.h
index c2c9844..b95e410 100644
--- a/libdrizzle-5.1/query.h
+++ b/libdrizzle-5.1/query.h
@@ -79,7 +79,7 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
  * function and needs to be freed by the application when finished with.
  *
  * This function does not escape '%' or '_', if you need to, call
- * drizzle_escape_str with `is_pattern` set to a non-zero value.
+ * drizzle_escape_str with `is_pattern` set to `true` value.
  *
  * @param[in] con a connection object
  * @param[in,out] to the destination string
@@ -97,7 +97,7 @@ ssize_t drizzle_escape_string(drizzle_st *con, char **to, const char *from, cons
  * This function escapes the following characters:
  * '\0' (0x00), '\'' (0x27), '"' (0x22), '\b' (0x08), '\n' (0x0A),
  * '\r' (0x0D), '\t' (0x09), '\Z' (0x26), '\\' (0x5C).
- * In case `is_pattern` is set to a non-zero, '%' (0x25) and '_' (0x5F)
+ * In case `is_pattern` is set to `true`, '%' (0x25) and '_' (0x5F)
  * will be escaped as well.
  *
  * The to parameter is allocated by the function and needs to be freed by
```

+ commit message

libdrizzle-5.1/query.h
```diff
@@ -88,6 +91,31 @@ drizzle_result_st *drizzle_query(drizzle_st *con,
...
+ * will be escaped as well.
+ *
+ * The to parameter is allocated by the function and needs to be freed by
+ * the application when finished with.
+ *
+ * @param[in] con a connection object
+ * @param[in,out] to the destination string
+ * @param[in] from the source string
+ * @param[in] from_size the length of the source string
+ * @param[in] is_pattern whether to escape '%' and '_'.
```
andreas-bok-sociomantic, line 110: nit picky, but could the alignment be like: 

```
 * @param[in] is_pattern whether to escape '%' and '_'. If set to `true`, they
 *                       will be escaped, so the string can be used in a LIKE
 *                       clause for example.
```

tests/unit/escape.c
```diff
@@ -52,17 +52,34 @@ int main(int argc, char* argv[])
 
   // Test for data too long
   char *out;
-  uint64_t out_len = drizzle_escape_string(NULL, &out, in, strlen(in));
+  uint64_t out_len = drizzle_escape_str(NULL, &out, in, strlen(in), 0);
```
andreas-bok-sociomantic: I guess `true/false` can be used as value for the `is_pattern` parameter now?

andreas-bok-sociomantic (commented):
Just a few comments :)

mathias-lang-sociomantic:
Addressed @andreas-bok-sociomantic 's comments

andreas-bok-sociomantic:
LGTM

PR #127, opened 2017-03-24
Remove reference to drizzle_clone from documentation

andreas-bok-sociomantic:
`drizzle_clone` was removed from the public API at an
earlier point but it wasn't reflected in the
documentation

The mention is removed from rst api docs and function
documentation in header files.

### DIFF
--- a/docs/api/binlog.rst
+++ b/docs/api/binlog.rst
@@ -172,6 +172,6 @@ Functions
 
    The memory (de)allocation of the filename buffer must be done by the client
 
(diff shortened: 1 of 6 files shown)


ben-palmer-sociomantic:
I assume there are no more "related functions" either? if so, LGTM

andreas-bok-sociomantic:
> I assume there are no more "related functions" either? if so, LGTM

Yes, the only way to create a drizzle instance is `drizzle_create`

andreas-bok-sociomantic:
tx

PR #128, opened 2017-03-29
Bump version to v5.6.0

andreas-bok-sociomantic:
New ABI version: 10.2.0


ben-palmer-sociomantic:
I think the commit message is wrong, otherwise LGTM.

PR #138, opened 2017-04-08
Fix: remove outdated symlink for libdrizzle-redux.so.10

andreas-bok-sociomantic:
The symlink part of the v5.4.0 release to not break
binary compatibility
It is no longer needed in v6.0.0


andreas-bok-sociomantic:
Duplicate PR

PR #139, opened 2017-04-09
Fix: remove outdated symlink for libdrizzle-redux.so.10

andreas-bok-sociomantic:
The symlink part of the v5.4.0 release to not break
binary compatibility
It is no longer needed in v6.0.0


deb/after_install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
+
+if [ -h /usr/lib/libdrizzle-redux.so.9 ]; then
+    echo "Removing outdated symlink"
+    sudo rm /usr/lib/libdrizzle-redux.so.10 /usr/lib/libdrizzle-redux.so.9 2>/dev/null
```
gautam-kotian-sociomantic, line 8: Do you need the `sudo` here? Won't the package be installed with superuser privileges anyway?

deb/after_install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
daniel-zullo, line 1: I think it is better to use `/bin/sh` (Bourne Shell) rather than bash for deb package after install script.

gautam-kotian-sociomantic, line 1: Luca once commented on a different packaging PR that after install scripts shouldn't be bash scripts as then we'd need to specify `bash` as a dependency in the packaging configuration. Can you change this to `#!/bin/sh` instead?

deb/after_install.sh
gautam-kotian-sociomantic: Presumably you used `2>/dev/null` to suppress errors if the sym links don't exist. If that was the only reason, I'd suggest using `rm -f` instead and avoid redirecting `stderr` to `/dev/null`.

deb/after_install.sh
```diff
@@ -0,0 +1,13 @@
...
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
+
+if [ -h /usr/lib/libdrizzle-redux.so.9 ]; then
+    echo "Removing outdated symlink"
+    sudo rm /usr/lib/libdrizzle-redux.so.10 /usr/lib/libdrizzle-redux.so.9 2>/dev/null
+    if [ -e "/usr/lib/libdrizzle-redux.so.9" ]; then
+        echo "Couldn't remove symlink '/usr/lib/libdrizzle-redux.so.9'"
+              "Please remove it manually"
```
gautam-kotian-sociomantic, line 11: Wouldn't you want to exit with a non-zero status code in this case?

deb/after_install.sh
andreas-bok-sociomantic: > Do you need the sudo here? Won't the package be installed with superuser privileges anyway?
   
true sudo is not needed 

> resumably you used 2>/dev/null to suppress errors if the sym links don't exist

Indeed, I will change to your suggestion

daniel-zullo, line 8: why `/usr/lib/libdrizzle-redux.so.10` is removed here?

deb/after_install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+#
+# Script which is run after install to ensure binary compatibility between
+# this and the previous version of libdrizzle-redux
+
+if [ -h /usr/lib/libdrizzle-redux.so.9 ]; then
+    echo "Removing outdated symlink"
+    sudo rm /usr/lib/libdrizzle-redux.so.10 /usr/lib/libdrizzle-redux.so.9 2>/dev/null
+    if [ -e "/usr/lib/libdrizzle-redux.so.9" ]; then
```
daniel-zullo, line 9: why is `-e` used here? (in the statement above `-h` was used)

deb/after_install.sh
andreas-bok-sociomantic: Yes, maybe adding the choice to remove the symlink if it exists is better, e.g

```sh
# if the symlink exists 
read -p "<Your Friendly Message here> : y/n/cancel" CONDITION;

if [ "$CONDITION" == "y" ]; then
   # do something here!
fi
```

deb/after_install.sh
daniel-zullo: Also you wouldn't need to use `sudo` as the after install runs with root privileges

daniel-zullo (commented):
I think we could take an action if the `rm` fails rather than checking again for the symlink in a nested if statement.

deb/after_install.sh
andreas-bok-sociomantic: > why /usr/lib/libdrizzle-redux.so.10 is removed here?

Right, only `so.9` should be removed :man_facepalming:

deb/after_install.sh
andreas-bok-sociomantic: Will change to `-h`

deb/after_install.sh
hatem-oraby-sociomantic: Some might want to install the package as part of automated install (using `apt-get install -y`), whatever choice you go with just make sure that your script plays nicely with `-y`.

hatem-oraby-sociomantic: If this package won't work properly if the file exists, then IMO you should quit installation.
I think just:
`test -e "/usr/lib/libdrizzle-redux.so.9"`
Should be enough to fail installation if the file still exists

andreas-bok-sociomantic:
Thanks for the comments. Given the issue #149 created by @mathias-lang-sociomantic it might collide with the installation of multiple versions of `libdrizzle-redux`. It would break an existing v5 installation when if the client is using the version with `so.9` libversion with the symlink. I will block it with the possibly resolve to close it without merging. 
Thanks again

andreas-bok-sociomantic:
Not relevant anymore after [v6.0.3](https://github.com/sociomantic-tsunami/libdrizzle-redux/releases/tag/v6.0.3)

PR #140, opened 2017-04-09
Improve error message for latex package check

andreas-bok-sociomantic:


ben-palmer-sociomantic:
LGTM

ben-palmer-sociomantic:
Its not technically a bug fix but I think it is ok to have it as part of this patch release.

PR #153, opened 2017-04-22
Bugfix: skip NULL fields when reading from MySQL

chenzhuoyu:
In `drizzle_state_binary_field_read()` at `field.cc:401`

`NULL` fields have no bytes but bit-masked by `con->result->null_bitmap`, we should skip them.

For example, given table:
```sql
CREATE TABLE `test_table` (
  a VARCHAR(10) DEFAULT NULL,
  b VARCHAR(10) DEFAULT NULL,
  c INT(10) DEFAULT NULL,
  d VARCHAR(10) DEFAULT NULL
) ENGINE=InnoDB;
```

and row

| a     | b      | c   | d     |
| ----- | ------ | --- | ----- |
| test1 | `NULL` | 123 | test2 |

Since `NULL` fields are not presented in the response body, if we try to read field `b` from the response, it's not a valid operation and may cause serious problems.

This patch fixes this problem by skipping `NULL` fields when reading from MySQL's response.

### DIFF
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -402,9 +402,21 @@ drizzle_return_t drizzle_state_binary_field_read(drizzle_st *con)
 {
   drizzle_return_t ret;
 
-  switch(con->result->column_buffer[con->result->field_current].type)
+  /* Skip NULL fields */
+  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] &
+         (1 << ((con->result->field_current_read + 2) % 8)))
+  {
+    con->result->field_current_read++;
+  }
+
+  switch(con->result->column_buffer[con->result->field_current_read].type)
   {
     case DRIZZLE_COLUMN_TYPE_NULL:
+      /* This should not happen since NULL fields are masked by `null_bitmap`, but cover it anyway */
+      drizzle_set_error(con, __func__,
+        "Unexpected column type: DRIZZLE_COLUMN_TYPE_NULL");
+      con->result->field_size= 0;
+      return DRIZZLE_RETURN_UNEXPECTED_DATA;
       break;
     case DRIZZLE_COLUMN_TYPE_TINY:
       con->result->field_size= 1;
@@ -462,6 +474,7 @@ drizzle_return_t drizzle_state_binary_field_read(drizzle_st *con)
   con->result->field_total= con->result->field_size;
 
   con->result->field_current++;
+  con->result->field_current_read++;
   con->pop_state();
   return DRIZZLE_RETURN_OK;
 }
--- a/libdrizzle/statement.cc
+++ b/libdrizzle/statement.cc
@@ -392,7 +392,6 @@ drizzle_return_t drizzle_stmt_fetch(drizzle_stmt_st *stmt)
   drizzle_return_t ret= DRIZZLE_RETURN_OK;
   uint16_t column_counter= 0, current_column= 0;
   drizzle_row_t row;
-  drizzle_column_st *column;
 
   if (stmt == NULL)
   {
@@ -422,11 +421,12 @@ drizzle_return_t drizzle_stmt_fetch(drizzle_stmt_st *stmt)
   {
     return DRIZZLE_RETURN_ROW_END;
   }
-  drizzle_column_seek(stmt->execute_result, 0);
 
   for (column_counter = 0; column_counter < stmt->execute_result->column_count; column_counter++)
   {
     drizzle_bind_st *param= &stmt->result_params[column_counter];
+    drizzle_column_st *column= drizzle_column_index(stmt->execute_result, column_counter);
+
     /* if this row is null in the result bitmap, skip first 2 bits */
     if (stmt->execute_result->null_bitmap[(column_counter+2)/8] & (1 << ((column_counter+2) % 8)))
     {
@@ -439,7 +439,6 @@ drizzle_return_t drizzle_stmt_fetch(drizzle_stmt_st *stmt)
       uint16_t short_data;
       uint32_t long_data;
       uint64_t longlong_data;
-      column= drizzle_column_next(stmt->execute_result);
       param->type= column->type;
       param->options.is_null= false;
       param->length= stmt->execute_result->field_sizes[current_column];
--- a/libdrizzle/result.h
+++ b/libdrizzle/result.h
@@ -65,6 +65,7 @@ struct drizzle_result_st
   uint64_t row_current;
 
   uint16_t field_current;         /* field index */
+  uint16_t field_current_read;    /* index of the field currently being read */
   uint64_t field_total;           /* total length of the field currently being read */
   uint64_t field_offset;          /* offset within field of most recently read field fragment (0 if first/only fragment) */
   uint32_t field_size;            /* size of most recently read field value or fragment of field value; max 2^24 */
@@ -102,6 +103,7 @@ struct drizzle_result_st
     row_count(0),
     row_current(0),
     field_current(0),
+    field_current_read(0),
     field_total(0),
     field_offset(0),
     field_size(0),
--- a/libdrizzle/row.cc
+++ b/libdrizzle/row.cc
@@ -303,6 +303,7 @@ drizzle_return_t drizzle_state_row_read(drizzle_st *con)
     con->result->row_count++;
     con->result->row_current++;
     con->result->field_current= 0;
+    con->result->field_current_read= 0;
   }
 
   con->pop_state();
--- a/tests/unit/include.am
+++ b/tests/unit/include.am
@@ -141,3 +141,8 @@ nodist_EXTRA_tests_unit_statement_char_SOURCES = dummy.cxx
 check_PROGRAMS+= tests/unit/statement_char
 noinst_PROGRAMS+= tests/unit/statement_char
 
+tests_unit_statement_nulls_SOURCES= tests/unit/statement_nulls.c tests/unit/common.c
+tests_unit_statement_nulls_LDADD= libdrizzle/libdrizzle-redux.la
+nodist_EXTRA_tests_unit_statement_nulls_SOURCES = dummy.cxx
+check_PROGRAMS+= tests/unit/statement_nulls
+noinst_PROGRAMS+= tests/unit/statement_nulls
--- a/tests/unit/statement_nulls.c
+++ b/tests/unit/statement_nulls.c
@@ -1,0 +1,103 @@
+/*  vim:expandtab:shiftwidth=2:tabstop=2:smarttab:
+ *
+ *  Drizzle Client & Protocol Library
+ *
+ * Copyright (C) 2012 Drizzle Developer Group
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *     * The names of its contributors may not be used to endorse or
+ * promote products derived from this software without specific prior
+ * written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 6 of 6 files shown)


libdrizzle/statement.cc
```diff
@@ -392,7 +392,6 @@ drizzle_return_t drizzle_stmt_fetch(drizzle_stmt_st *stmt)
   drizzle_return_t ret= DRIZZLE_RETURN_OK;
   uint16_t column_counter= 0, current_column= 0;
   drizzle_row_t row;
-  drizzle_column_st *column;
```
andreas-bok-sociomantic, line 395: Do we need to move the declaration of the `drizzle_column_st ` into the for loop?

libdrizzle/field.cc
```diff
@@ -402,9 +402,17 @@ drizzle_return_t drizzle_state_binary_field_read(drizzle_st *con)
 {
   drizzle_return_t ret;
 
-  switch(con->result->column_buffer[con->result->field_current].type)
+  /* Skip NULL fields */
+  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] & (1 << ((con->result->field_current_read + 2) % 8)))
```
andreas-bok-sociomantic: Just understand. Is the added offset of 2 in `con->result->field_current_read + 2` a consequence of the cautionary note in the MySQL internals doc for NULL-Bitmap?, cf. https://dev.mysql.com/doc/internals/en/null-bitmap.html

andreas-bok-sociomantic: Please format as:

```c++
  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] &
         (1 << ((con->result->field_current_read + 2) % 8)))
```

andreas-bok-sociomantic (commented):
Hi @chenzhuoyu, thanks for your contribution :)
LGTM just have a few comments. 
Just to clarify. Is the changes based on the MySQL doc for NULL-bitmap?  https://dev.mysql.com/doc/internals/en/null-bitmap.html. 

Secondly, we haven't encountered the problems with would this bug could have caused. We read data with NULL values which is parsed by the current implementation. 
Is it possible to make a test case which would reproduce the bug?

libdrizzle/field.cc
chenzhuoyu: 1. Yes, and I've captured the actual packet on wire using WireShark, confirmed that 2-bits offset is actually there, but in fact I have no idea why, and I found nothing about this magic number in that manual either :(  Maybe they want to maintain protocol compatibility ?

2. OK, I'll do it tomorrow :)

libdrizzle/statement.cc
chenzhuoyu: Not necessarily, but it looks a bit nicer

chenzhuoyu:
We found this bug by running unittests in our project, for some reason I cannot provide the source code, but I will construct a seperate test program to reproduce this problem :)

chenzhuoyu:
.....Oops I clicked the wrong button OMG .....

chenzhuoyu:
Here I constructed a simple program to perfrom the test:
[Test case, on my Gist](https://gist.github.com/chenzhuoyu/70554377c5e82a78bae4becc31e0873b)

g++ version:
```
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Linux version:
```
Linux 4.4.0-31-generic #50-Ubuntu SMP Wed Jul 13 00:07:12 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

MySQL version:
```
mysql  Ver 14.14 Distrib 5.7.17, for Linux (x86_64) using  EditLine wrapper
```

libdrizzle-redux configure command:
```bash
./configure --prefix=/usr/local --enable-shared=no --enable-static=yes
```

libdrizzle-redux build command:
```bash
make -j24
make install
```

Test case build command:
```bash
g++ -g -Wall -pthread -o test test.cpp -ldrizzle-redux -lssl -lcrypto -lz
```

Table creation SQL:
```sql
CREATE TABLE `test_table` (
    `id`        int(10)         NOT NULL AUTO_INCREMENT,
    `word`      varchar(255)    NOT NULL,
    `time`      varchar(16)     NOT NULL,
    `uid`       varchar(32)     DEFAULT NULL,
    `question`  varchar(255)    DEFAULT NULL,
    `answer`    varchar(255)    DEFAULT NULL,
    `is_valid`  int(1)          NOT NULL DEFAULT 1,
    `source`    varchar(32)     DEFAULT NULL,
    `type`      int(2)          DEFAULT NULL,
    `uuid`      char(36)        DEFAULT NULL,
    PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

with content:
```sql
INSERT INTO `test_table` (`word`, `time`, `question`, `answer`, `is_valid`, `source`, `uuid`)
VALUES ('test word', '1492672507', 'test question', 'test answer', 0, 'x', 'd7f15abb-a482-43be-99c2-c175408c1bd9');
```

MySQL client result:
```
mysql> SELECT * FROM test_table;
+----+-----------+------------+------+---------------+-------------+----------+--------+------+--------------------------------------+
| id | word      | time       | uid  | question      | answer      | is_valid | source | type | uuid                                 |
+----+-----------+------------+------+---------------+-------------+----------+--------+------+--------------------------------------+
|  1 | test word | 1492672507 | NULL | test question | test answer |        0 | x      | NULL | d7f15abb-a482-43be-99c2-c175408c1bd9 |
+----+-----------+------------+------+---------------+-------------+----------+--------+------+--------------------------------------+
1 row in set (0.00 sec)
```

My result of this test case:
```
$ ./test
execute
buffer
** FATAL : [9] drizzle_state_packet_read:bad packet number:14:0
```

Also the capture file is attached here:
[capture.zip](https://github.com/sociomantic-tsunami/libdrizzle-redux/files/960839/capture.zip)

chenzhuoyu:
@andreas-bok-sociomantic hope this helps :)

chenzhuoyu:
@bokchan  I've updated the code format

andreas-bok-sociomantic:
Great with the test case for the bug. 
The PR looks good to me, so just a matter of me running the test case on my computer. 
Get back to you soon

andreas-bok-sociomantic:
Hi @chenzhuoyu, ran the testcase you provided, And indeed it fails when buffering a prepared statement query. 
I created a unittest from your example and pushed it to this branch https://github.com/andreas-bok-sociomantic/libdrizzle-redux/tree/skip-null-fields
You are welcome to add the [commit](https://github.com/andreas-bok-sociomantic/libdrizzle-redux/commit/909f8090178564658f4bb14cbc75b677c20d58f9) with the added test case to this PR. 
Otherwise I will just make a new PR with that addition. 

Thanks again for your great work :)

andreas-bok-sociomantic:
ping @chenzhuoyu

chenzhuoyu:
I've cherry picked the test-case commit from your repository @andreas-bok-sociomantic 
Sorry for the late reply, I'v been on vacation these days and did not have chance to read your email before.

andreas-bok-sociomantic:
Cool, no worries. I will just take one more look. But it should ready to merge :)

libdrizzle/field.cc
```diff
@@ -402,9 +402,18 @@ drizzle_return_t drizzle_state_binary_field_read(drizzle_st *con)
...
+  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] &
+         (1 << ((con->result->field_current_read + 2) % 8)))
+  {
+    con->result->field_current_read++;
+  }
+
+  switch(con->result->column_buffer[con->result->field_current_read].type)
   {
     case DRIZZLE_COLUMN_TYPE_NULL:
+      /* This should not happen since NULL fields are masked by `null_bitmap`, but cover it anyway */
```
andreas-bok-sociomantic, line 415: A colleague, @daniel-zullo-sociomantic , noted that if the unlikely case should happen and this case is hit, `field_current_read` would be incorrectly updated twice. 
How do you feel about this instead, which would indeed make it impossible to parse a field with a null value:

```cpp 
  /* Skip NULL fields */
  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] &
         (1 << ((con->result->field_current_read + 2) % 8)))
  {
    con->result->field_current_read++;
  }

  drizzle_column_type_t column_type =
    con->result->column_buffer[con->result->field_current_read].type;

  assert(column_type !=  DRIZZLE_COLUMN_TYPE_NULL);

  switch(column_type)
  {
    case DRIZZLE_COLUMN_TYPE_TINY:
      con->result->field_size= 1;
      break;
  // ...
  }
```

andreas-bok-sociomantic: This might be better since it won't halt the program and will allow the user to handle the error and get an error message 

```cpp
drizzle_return_t drizzle_state_binary_field_read(drizzle_st *con)
...
  /* Skip NULL fields */
  while (con->result->null_bitmap[(con->result->field_current_read + 2) / 8] &
         (1 << ((con->result->field_current_read + 2) % 8)))
  {
    con->result->field_current_read++;
  }

  drizzle_column_type_t column_type =
    con->result->column_buffer[con->result->field_current_read].type;

  switch(column_type)
  {
    case DRIZZLE_COLUMN_TYPE_NULL:
      /* This should not happen since NULL fields are masked by `null_bitmap`,
       * but cover it anyway */
      drizzle_set_error(con, __func__,
        "Unexpected column type: DRIZZLE_COLUMN_TYPE_NULL");
      con->result->field_size= 0;
      return DRIZZLE_RETURN_UNEXPECTED_DATA;
      break;
    case DRIZZLE_COLUMN_TYPE_TINY:
...
```

andreas-bok-sociomantic:
ping @chenzhuoyu

chenzhuoyu:
Sorry for the late reply, but I can't figure out in which case `field_current_read` could be updated twice :(

Actually I don't even think `column_type` could be `DRIZZLE_COLUMN_TYPE_NULL` ever.

I've gone through the source (by `fgrep -nR 'DRIZZLE_COLUMN_TYPE_NULL' .`), and found there were only 2 places that sets something to `DRIZZLE_COLUMN_TYPE_NULL`, which are:

1. The first one is for setting query parameter types in prepared statements, thus have nothing to do with result column types.

```cpp
/* see statement_param.cc:134 in "drizzle_stmt_set_null" */
/* 132 */ drizzle_return_t drizzle_stmt_set_null(drizzle_stmt_st *stmt, uint16_t param_num)
/* 133 */ {
/* 134 */   return drizzle_stmt_set_param(stmt, param_num, DRIZZLE_COLUMN_TYPE_NULL, NULL, 0, false);
/* 135 */ }
```

which invokes `drizzle_stmt_set_param`, at line `58` it sets the query parameter type

```cpp
/* see statement_param.cc:58 in "drizzle_stmt_set_param" */
/* 46 */ drizzle_return_t drizzle_stmt_set_param(drizzle_stmt_st *stmt, uint16_t param_num, drizzle_column_type_t type, const void *data, size_t length, bool is_unsigned)
/* 47 */ {
/* 48 */   if ((stmt == NULL) || (param_num >= stmt->param_count))
/* 49 */   {
/* 50 */     return DRIZZLE_RETURN_INVALID_ARGUMENT;
/* 51 */   }
/* 52 */   if (stmt->state < DRIZZLE_STMT_PREPARED)
/* 53 */   {
/* 54 */     drizzle_set_error(stmt->con, __func__, "stmt object has not been prepared");
/* 55 */     return DRIZZLE_RETURN_STMT_ERROR;
/* 56 */   }
/* 57 */ 
/* 58 */   stmt->query_params[param_num].type= type;
/* 59 */   stmt->query_params[param_num].data= (void*)data;
/* 60 */   stmt->query_params[param_num].length= length;
/* 61 */   stmt->query_params[param_num].options.is_unsigned= is_unsigned;
/* 62 */   stmt->query_params[param_num].is_bound= true;
/* 63 */ 
/* 64 */   return DRIZZLE_RETURN_OK;
/* 65 */ }
```

2. The second one happens **AFTER** the row been buffered, also have nothing to do with result column types.

```cpp
/* see statement.cc:435 in "drizzle_stmt_fetch" */
/* 410 */ /* ... */
/* 411 */ if (stmt->execute_result->options & DRIZZLE_RESULT_BUFFER_ROW)
/* 412 */ {
/* 413 */   row= drizzle_row_next(stmt->execute_result);
/* 414 */ }
/* 415 */ else
/* 416 */ {
/* 417 */   row= drizzle_row_buffer(stmt->execute_result, &ret);
/* 418 */ }
/* 419 */ 
/* 420 */ if (row == NULL)
/* 421 */ {
/* 422 */   return DRIZZLE_RETURN_ROW_END;
/* 423 */ }
/* 424 */ 
/* 425 */ for (column_counter = 0; column_counter < stmt->execute_result->column_count; column_counter++)
/* 426 */ {
/* 427 */   drizzle_bind_st *param= &stmt->result_params[column_counter];
/* 428 */   drizzle_column_st *column= drizzle_column_index(stmt->execute_result, column_counter);
/* 429 */ 
/* 430 */   /* if this row is null in the result bitmap, skip first 2 bits */
/* 431 */   if (stmt->execute_result->null_bitmap[(column_counter+2)/8] & (1 << ((column_counter+2) % 8)))
/* 432 */   {
/* 433 */     param->options.is_null= true;
/* 434 */     param->length= 0;
/* 435 */     param->type= DRIZZLE_COLUMN_TYPE_NULL;
/* 436 */   }
/* 437 */   else
/* 438 */ /* ... */
```

At line `435` it sets `param->type` to `DRIZZLE_COLUMN_TYPE_NULL`, but this happens **AFTER** both `drizzle_row_next` and `drizzle_row_buffer` (see `statement.cc:413 and :417 in "drizzle_stmt_fetch"`)

But anyway, I agree with you that adding an error message in that "case" statement is a good idea, although I don't think it might ever happens [Facepalm]

chenzhuoyu:
@andreas-bok-sociomantic

andreas-bok-sociomantic:
> But anyway, I agree with you that adding an error message in that "case" statement is a good idea, although I don't think it might ever happens [Facepalm]

Thanks for looking into it again. I suggest based on your comments, that covering the case (which in fact cannot happen) and setting an error message is sufficient. If not else to have the switch handle all 
`drizzle_column_type_t` enum values. 

With that change I would be happy to merge

```cpp
     case DRIZZLE_COLUMN_TYPE_NULL:
      /* This should not happen since NULL fields are masked by `null_bitmap`,
       * but cover it anyway */
      drizzle_set_error(con, __func__,
        "Unexpected column type: DRIZZLE_COLUMN_TYPE_NULL");
      con->result->field_size= 0;
      return DRIZZLE_RETURN_UNEXPECTED_DATA;
      break;
```

chenzhuoyu:
@andreas-bok-sociomantic I've added these changes

andreas-bok-sociomantic:
Cool thanks

Issue #155, opened 2017-04-29
Compile error for unsupported socket option TCP_KEEPIDLE


andreas-bok-sociomantic:
The #114 broke the build for systems not supporting **TCP_KEEPIDLE** for `setsockopt`,
 
`libdrizzle/conn.cc:1653:41: error: 'TCP_KEEPIDLE' was not declared in this scope
   ret= setsockopt(con->fd, IPPROTO_TCP, TCP_KEEPIDLE`

A check for support should be added so `setsockopt` with **TCP_KEEPIDLE** is only when it is supported

PR #156, opened 2017-04-29
Bugfix: compile error for unsupported socket option TCP_KEEPIDLE

andreas-bok-sociomantic:
Fix the issue reported in #155 
Also sanitizes the use preprocessor checks for _WIN32 in the internal `_setsockopt` function.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -121,6 +121,9 @@ AC_CHECK_HEADERS([windows.h])
 AC_CHECK_HEADERS([winsock2.h])
 AC_CHECK_HEADERS([ws2tcpip.h])
 
+# Check for setsockopt support
+AX_CHECK_SETSOCKOPT
+
 # Checks for typedefs, structures, and compiler characteristics.
 AC_CHECK_TYPES([in_port_t])
 AC_TYPE_SSIZE_T
--- a/libdrizzle/conn.cc
+++ b/libdrizzle/conn.cc
@@ -1634,7 +1634,7 @@ static drizzle_return_t _setsockopt(drizzle_st *con)
   ret = 1;
 #ifdef _WIN32
   ret= setsockopt(con->fd, SOL_SOCKET, SO_KEEPALIVE, (const char*)&ret, optlen_int);
-#else
+#elif defined HAVE_SO_KEEPALIVE
   ret= setsockopt(con->fd, SOL_SOCKET, SO_KEEPALIVE,&ret, optlen_int);
 #endif /* _WIN32 */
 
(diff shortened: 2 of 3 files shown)


m4/ax_check_socket_options.m4
```diff
@@ -0,0 +1,186 @@
...
+dnl     TCP_KEEPIDLE
+dnl     TCP_KEEPINTVL
+dnl
+dnl The macro is a modified version of the acinclude.m4 from the ZeroMQ library
+dnl http://github.com/zeromq/libzmq
+dnl @license Mozilla Public License v2 (MPLv2)
+
+
+dnl ################################################################################
+dnl # CHECK_SO_KEEPALIVE([action-if-found], [action-if-not-found])          #
```
daniel-zullo: Documentation headers looks wrong. The `#` might be moved to the end of the line. If my comment applies then you should fix the other documentation headers too.

don-clugston-sociomantic:
LGTM, except for the m4 macro which I don't understand well enough to review easily. Is it copied from somewhere? If it is, I think the commit should say so.

m4/ax_check_socket_options.m4
```diff
@@ -0,0 +1,186 @@
...
+    );
+}
+    ],
+    [AC_MSG_RESULT(yes) ; cv_tcp_keepalive="yes" ; $1],
+    [AC_MSG_RESULT(no)  ; cv_tcp_keepalive="no"  ; $2],
+    [AC_MSG_RESULT(not during cross-compile) ; cv_tcp_keepalive="no"]
+    )
+}])
+
+AC_DEFUN([AX_SOCKET_OPTIONS],
```
daniel-zullo: What about rename it to `AX_CHECK_SOCKET_OPTIONS`?

m4/ax_check_socket_options.m4
```diff
@@ -0,0 +1,186 @@
...
+            [1],
+            [Whether TCP_KEEPINTVL is supported.])
+        ])
+
+    CHECK_TCP_KEEPALIVE([
+        AC_DEFINE([HAVE_TCP_KEEPALIVE],
+            [1],
+            [Whether TCP_KEEPALIVE is supported.])
+        ])
+
```
daniel-zullo: you might want to remove this empty line

daniel-zullo (commented):
Just a couple of minor comments and LGTM

andreas-bok-sociomantic:
> LGTM, except for the m4 macro which I don't understand well enough to review easily. Is it copied from somewhere? If it is, I think the commit should say so.

Sure. I added the origin to the m4 macro header. Will add it to the ci msg as well

andreas-bok-sociomantic:
In addition added cross-platform checks in each of the test programs

andreas-bok-sociomantic:
Renamed the m4 macro to `ax_check_setsockopt.m4`

ben-palmer-sociomantic:
LGTM

ben-palmer-sociomantic:
Have we tested that the socket options are still set under linux?

andreas-bok-sociomantic:
> Have we tested that the socket options are still set under linux?

Yes

Issue #62, opened 2016-09-09
Compile error with clang on osx for long long 


andreas-bok-sociomantic:
```
Apple LLVM version 7.3.0 (clang-703.0.31)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
```

Running `./configure` gives the following error

``` sh
-DPIC -o libdrizzle/.libs/libdrizzle_libdrizzle_redux_la-field.o
libdrizzle/field.cc:150:5: error: 'long long' is a C++11 extension [-Werror,-Wc++11-long-long]
#if SIZE_MAX < UINT64_MAX
    ^
/usr/include/stdint.h:153:20: note: expanded from macro 'SIZE_MAX'
#define SIZE_MAX          UINT64_MAX
                          ^
/usr/include/stdint.h:87:27: note: expanded from macro 'UINT64_MAX'
#define UINT64_MAX        18446744073709551615ULL
                          ^
libdrizzle/field.cc:150:16: error: 'long long' is a C++11 extension [-Werror,-Wc++11-long-long]
#if SIZE_MAX < UINT64_MAX
               ^
/usr/include/stdint.h:87:27: note: expanded from macro 'UINT64_MAX'
#define UINT64_MAX        18446744073709551615ULL
                          ^
2 errors generated.
make[1]: *** [libdrizzle/libdrizzle_libdrizzle_redux_la-field.lo] Error 1
make: *** [all] Error 2
```

The code in question is : 

``` cxx
libdrizzle/field.cc:150

#if SIZE_MAX < UINT64_MAX
  if (wire_size >= SIZE_MAX)
  {
    drizzle_set_error(result->con, __func__, "Field is larger than memory.");
    *ret_ptr= DRIZZLE_RETURN_MEMORY;
    return NULL;
  }
#endif
```

The origin of the bug seems to be this commit 42823bf941c26c6b3a36f83b4e33abd207e3e57c, after which field sizes and offsets were stored as`uint64_t` 

Possible fixes are: 
1.  Remove `_APPEND_COMPILE_FLAGS_ERROR([-Wlong-long])` from compile flags, i.e. removing this line from [ax_harden_compiler_flags.m4#L205](https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/master/m4/ax_harden_compiler_flags.m4#L205) 
2. If the target platform is darwin and the compiler is clang add `-std=c++11 -stdlib=libc++` to `CXXFLAGS`

[...]

PR #159, opened 2017-04-29
Bugfix: Fix check for c++11 support and long-long compiler warning 

andreas-bok-sociomantic:
Letting autoconf include AX_CXX_COMPILE_STDCXX
automatically to check the C++ compiler for c++11 support
did not work correctly, cf. 
http://stackoverflow.com/questions/24923600/ax-cxx-compile-stdcxx-11-doesnt-work-syntax-error-near-unexpected-token 

Thus the file ax_cxx_compile_stdcxx.m4 has been added

In addition std-gnu11.m4 has been added which redefine
AC_PROG_CC with check for C-compiler support for gnu11

Fixes the error which was reported in #62:
error: 'long long' is a C++11 extension [-Werror,-Wc++11-long-long]

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,6 +61,28 @@ Attention!
   We apologize for any inconveniences and promise to look into the issue as soon
   as possible.
 
+**Supported compilers**
+
+.. csv-table::
+  :header: "Compiler","Version"
+
+   gcc, >=4.8.x
+   llvm clang, >=3.3
+   apple clang, >=5.0 (clang-500.2.75)
+
+**Other Dependencies**
+
+.. csv-table::
+   :header: "Dependency", "Version"
+
+   autoconf, >=2.61
+   zlib1g-dev, >=0.13.x
+   libtool, >=2.x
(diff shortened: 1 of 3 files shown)


andreas-bok-sociomantic:
Rebased on latest v5.5.x

ben-palmer-sociomantic:
I don't know too much about these changes but the `m4` files added are pretty in depth aren't they?

andreas-bok-sociomantic:
> I don't know too much about these changes but the m4 files added are pretty in depth aren't they?

Indeed. The `std-gnu11.m4` macro which checks for C compiler gnu11 compatibility might be overkill, since we anyway compile all as cpp code. I will test it

andreas-bok-sociomantic:
Updated
- removed `std-gnu11.m4`

ben-palmer-sociomantic:
Needs to be rebased on master

andreas-bok-sociomantic:
Rebased on latest v5.5.x

ben-palmer-sociomantic:
LGTM

PR #164, opened 2017-05-02
Add message that MinGw build is broken

andreas-bok-sociomantic:
Updates the READ and compiling docs with
message about broken MinGW build


ben-palmer-sociomantic:
LGTM

PR #162, opened 2017-04-30
Add OSX build to travis ci

andreas-bok-sociomantic:
Modifies the travis build configuration to include build on osx with clang

**Not open for review yet**

PR #166, opened 2017-05-11
Add RPM package build to travis ci script

andreas-bok-sociomantic:
Add a Docker build file and set up the
travis config to build the image

Run the container and build the rpm package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,19 +13,47 @@ dist: trusty
 
 services:
  - mysql
+ - docker
 
-## The following environment variables are required for
-#  running unittests
-#  Environment variables can be defined in the repository, but
-#  are overridden by values in .travis.yml.
+## Environment variables
 #
-#  Please refer to the section 'script' for info on
-#  SKIP_TEST_ON_ERROR
+#  The 'global' section defines variables which are added to every row in the
+#  build matrix, i.e. every build configuration.
+#  In turn, variables defined in the 'matrix' section each triggers a separate
+#  build.
+#
+#  Please note that test fails on first error as default
+#  To skip errors and fail silently set environment variable:
+#    SKIP_TEST_ON_ERROR=1
 env:
-  - MYSQL_SERVER=127.0.0.1 \
-    MYSQL_PASSWORD='' \
-    MYSQL_USER=root \
-    SKIP_TEST_ON_ERROR=0
+  global:
+    - MYSQL_SERVER=127.0.0.1
+      MYSQL_PASSWORD=''
+      MYSQL_USER=root
+      MYSQL_PORT=3306
+      DIST_PACKAGE_TARGET=DEB
+      SKIP_TEST_ON_ERROR=0
+  matrix:
+    - DIST_PACKAGE_TARGET=DEB
+    - DIST_PACKAGE_TARGET=RPM
+
+## Build matrix
+#
+#  The 'os', 'compiler' and 'env' settings are expanded into 3 jobs:
+#
+#  | os     | compiler | DIST_PACKAGE_TARGET |
+#  |--------|----------|---------------------|
+#  | linux  | clang    | DEB                 |
+#  | linux  | gcc      | DEB                 |
+#  | linux* | gcc      | RPM                 |
+#
+#  * the RPM package is built in a docker container running centos7
+matrix:
+  fast_finish: true
+  exclude:
+    - os: linux
+      compiler: clang
+      env: DIST_PACKAGE_TARGET=RPM
 
 ## Defines packages to be installed by apt
 #  2016/12/01 - mysql-5.6 was added to the default stack
@@ -36,31 +64,14 @@ env:
 #  apt:
 #    packages:
 
+## Install dependencies
 before_install:
-  - gem install fpm
-  - mysql -u root -e "SET PASSWORD FOR 'root'@'localhost' = PASSWORD('')"
+  - ./travis_configure.sh before_install
 
-## This script runs before the compiling
-#  1. Stop MySQL server
-#  2. Enable replication in mysql config file
-#  3. Start MySQL server
-#  4. Print mysql config for debug purposes
+## Prepare build system for tests
 before_script:
-  - sudo sed  -i 's/\[mysqld\]/[mysqld]\nserver-id = 1\nlog-bin = mysqlbin\nbinlog-format = ROW\n/g' /etc/mysql/my.cnf
-  - sudo service mysql restart
-  - mysqld --print-defaults
+  - ./travis_configure.sh before_script
 
-## Build steps
-#  - compile project
-#  - run all tests with 'make check'
-#
-#  Please note that test fails on first error as default
-#  To skip errors and fail silently set environment variable:
-#    SKIP_TEST_ON_ERROR=1
+## Run tests, and packaging scripts
 script:
-  - autoreconf -fi
-  - ./configure
-  - make
-  - make check
-  - make deb
-
+  - ./travis_configure.sh run_tests
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,34 @@
+# Docker build file to create an image running centos7 with MySQL and build
+# tools installed
+
+FROM centos/mysql-56-centos7:latest
+MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
+
+# Set environment variables
+ENV MYSQL_USER=user \
+    MYSQL_PASSWORD=pass \
+    MYSQL_ROOT_PASSWORD='' \
+    MYSQL_SERVER=127.0.0.1 \
+    MYSQL_PORT=3306 \
+    MYSQL_DATABASE=test \
+    MYSQL_MASTER_USER=user \
+    MYSQL_MASTER_PASSWORD=pass \
+    MYSQL_BINLOG_FORMAT=ROW \
+    SKIP_TEST_ON_ERROR=0
+
+# Change user to `root` and install required packages to build libdrizzle-redux.
+USER root
+RUN yum install -y deltarpm
+RUN yum install -y rpm-build zlib-devel libtool openssl \
+    openssl-devel automake autoconf gcc-c++
+
+# Add host directory to container
+ADD ./ /home/
+
+WORKDIR /home
+
+# Change back user to default user for the image: mysql
+USER mysql
+
+# Run the MySQL server with replication enabled
+CMD [ "run-mysqld-master" ]
--- a/travis_configure.sh
+++ b/travis_configure.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+# Script for setting up OS specific CI test environment on travis-ci.org
+#
+# Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+set -e
+
+# Script to print error message with the build is and environment variable
+# DIST_PACKAGE_TARGET
+#
+# $1 error message to show
+#
+# Returns 0
+print_error_msg ()
+{
+    test ! -z "$1" || exit 1
+    echo "$1"
+    echo "  os:      $TRAVIS_OS_NAME"
+    echo "  package: $DIST_PACKAGE_TARGET"
+
+    return 0
+}
+
+# Script which is run before the installation script is called
+#
+# Installs fpm on needed for deb packaging
+#
+# Other build configurations are not handled in this function
+#
+# Returns 0
+before_install()
+{
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            gem install fpm
+        fi
+    fi
+
+    return 0
+}
+
+# Enable binary logging by setting log_bin and server_id in the [mysqld]
+# section of the mysql config file
+#
+# $1 path to mysql config file
+#
+# Returns 0
+enable_mysqlbinlog()
+{
+    sudo sed -i -e 's/^# *server.id.*$/server_id=1/g' $1
+    sudo sed -i -e 's/^# *log.bin.*$/log_bin=mysql-bin/g' $1
+
+    return 0
+}
+
+# Invoked by the travis hook before_script to prepare the build environment
+#
+# Returns 0 or 1 if called with an invalid build configuration
+before_script()
+{
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            # Ensures MySQL is started with binlog enabled. This is done by
+            # modifying the default my.cnf file.
+            #  1. Enable replication in mysql config file
+            #  2. Restart MySQL server
+            #  3. Set root password to an empty string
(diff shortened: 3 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Docker build file to create a centos image for building rpm packages
+
+FROM centos:7
+MAINTAINER andreas-bok-sociomantic <andreas-bok@sociomantic.com>
```
mathias-lang-sociomantic: Isn't it `andreas.bok` ? Maybe we could have a public address for the project(s) ?

docker/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN yum install -y rpm-build zlib-devel libtool openssl \
+    openssl-devel automake autoconf gcc-c++ && \
+    yum clean all && \
+    rm -r -f /var/cache/*
```
mathias-lang-sociomantic: Y ?

rpm/include.am
```diff
@@ -29,6 +29,14 @@ clean-rpm:
 
 rpm: rpm-build
 
+docker-rpm:
+	@sed -i 's/%{__make} check/#%{__make} check/g' "rpm/spec.in"
```
mathias-lang-sociomantic: This could use a comment as well. Why are we disabling this check ?

.travis.yml
```diff
@@ -63,4 +64,6 @@ script:
   - make
   - make check
   - make deb
+  - docker build -f docker/Dockerfile -t centos:7 .
+  - docker run --rm -v $PWD:/home centos:7 make docker-rpm
```
mathias-lang-sociomantic: Sounds like this could be hugely simplified if you used a [build matrix](https://docs.travis-ci.com/user/customizing-the-build#Build-Matrix).

docker/Dockerfile
andreas-bok-sociomantic: You you mean `yum install -Y`? 

if so, **no**: 

```
yum help install
...
  -y, --assumeyes       Automatically answer yes for all question
```

.travis.yml
andreas-bok-sociomantic: Using build matrix is already part of [#162](https://github.com/sociomantic-tsunami/libdrizzle-redux/pull/162/files#diff-354f30a63fb0907d4ad57269548329e3R51)

rpm/include.am
andreas-bok-sociomantic: Basically to avoid setting up a MySQL server in the docker container. The check test is already done [here](https://github.com/andreas-bok-sociomantic/libdrizzle-redux/blob/dc05295f05bdd3c3cd8b572e14e3ce7bd441f4ce/.travis.yml#L65)

And it reduces build time

andreas-bok-sociomantic:
Updated based on comments from @mathias-lang-sociomantic 

```diff
diff --git a/.travis.yml b/.travis.yml
index c9b909d..b2e3839 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,4 +66,3 @@ script:
   - make deb
   - docker build -f docker/Dockerfile -t centos:7 .
   - docker run --rm -v $PWD:/home centos:7 make docker-rpm
-
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 2d7c21a..fe113b3 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,7 @@
 # Docker build file to create a centos image for building rpm packages
 
 FROM centos:7
-MAINTAINER andreas-bok-sociomantic <andreas-bok@sociomantic.com>
+MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
 
 RUN yum install -y rpm-build zlib-devel libtool openssl \
     openssl-devel automake autoconf gcc-c++ && \
diff --git a/rpm/include.am b/rpm/include.am
index cd160a5..c1f2f7f 100644
--- a/rpm/include.am
+++ b/rpm/include.am
@@ -35,6 +35,10 @@ docker-rpm:
 	$(SHELL) ./configure
 	$(MAKE) rpm
 	$(MAKE) clean-rpm
+	# Disable running the unittests in the %check section. The check is done
+	# previously and it is obsolete for checking that the RPM package is built
+	# without errors. Further more it reduces the build time since it isn't
+	# necessary to set up a MySQL server in the docker container
 	@sed -i 's/#%{__make} check/%{__make} check/g' "rpm/spec.in"
 
 release: rpm-sign

```

docker/Dockerfile
mathias-lang-sociomantic: No, I mean "Why" :P
Why do you `rm -rf /var/cache/*` ?

andreas-bok-sociomantic: Ah. See your point since the container is transient... :)

.travis.yml
mathias-lang-sociomantic: Yeah I reviewed this one later. IMO the approach you have here is complicated by the fact you are not using build matrix. I would strongly recommend to go with the build matrix, then update this P.R. based on it.

docker/Dockerfile
andreas-bok-sociomantic: Ah. See your point since the container is transient... :)

`yum clean all && \` can be omitted as well

.travis.yml
andreas-bok-sociomantic: >  IMO the approach you have here is complicated by the fact you are not using build matrix. I would strongly recommend to go with the build matrix, then update this P.R. based on it.

Wrt. the RPM build to the travis script I am not sure much is gained by using the matrix. Building the Dockerfile and running the container still needs to be done explicitly in the `script` section. 
I can integrate the matrix into this PR and move the RPM build commands to the aux script as I have done in #162, so it is prepared for merging with #162 but again IMHO solely wrt to the RPM build it won't make much of a difference

andreas-bok-sociomantic:
Updated travis config file to use 'build matrix' to configure jobs.

docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    MYSQL_DATABASE=test \
+    MYSQL_MASTER_USER=user \
+    MYSQL_MASTER_PASSWORD=pass \
+    MYSQL_BINLOG_FORMAT=ROW \
+    SKIP_TEST_ON_ERROR=0
+
+# Change user to `root` and install required packages to build libdrizzle-redux.
+USER root
+RUN yum install -y deltarpm
+RUN yum install -y rpm-build zlib-devel libtool openssl \
```
daniel-zullo, line 22: I guess you could have included `deltarpm` here and delete the line above. Or am I missing something?

travis_configure.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
```
daniel-zullo: Usually `sh` is used for package configuration/installation. Is there any special reason why `bash` is needed?

travis_configure.sh
```diff
@@ -0,0 +1,96 @@
...
+# Enable binary logging by setting log_bin and server_id in the [mysqld]
+# section of the mysql config file
+#
+# $1 path to mysql config file
+#
+# Returns 0
+function enable_mysqlbinlog
+{
+    sudo sed -i -e 's/^# *server.id.*$/server_id=1/g' $1
+    sudo sed -i -e 's/^# *log.bin.*$/log_bin=mysql-bin/g' $1
```
daniel-zullo: Do you really need `sudo` here?

travis_configure.sh
```diff
@@ -0,0 +1,96 @@
...
+# Returns 0
+function before_script {
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            # Ensures MySQL is started with binlog enabled. This is done by
+            # modifying the default my.cnf file.
+            #  1. Enable replication in mysql config file
+            #  2. Restart MySQL server
+            enable_mysqlbinlog "/etc/mysql/my.cnf"
+            sudo service mysql restart
```
daniel-zullo: do you also need to use `sudo` here?

docker/Dockerfile
andreas-bok-sociomantic: What do you mean?
I wanted to install deltarpm first, so the less stuff has to be downloaded when installing the next packages

travis_configure.sh
andreas-bok-sociomantic: No particular reason, will change.

travis_configure.sh
andreas-bok-sociomantic: I will try to remove and see if it passes

docker/Dockerfile
daniel-zullo: I meant that you could have merged the two commands into one.

andreas-bok-sociomantic:
> I meant that you could have merged the two commands into one.

Yes, but the purpose of first installing deltarpm is that yum uses it to only download the diff of a update. So it needs that app to be installed to perform that action when installing the next packages

docker/Dockerfile
daniel-zullo: > Yes, but the purpose of first installing deltarpm is that yum uses it to only download the diff of a update. So it needs that app to be installed to perform that action when installing the next packages

ok, got it

travis_configure.sh
andreas-bok-sociomantic: Nope, sudo is required to write to the root folder `/etc`. it is required for calling `service`

andreas-bok-sociomantic:
Feel free to merge, if there are no further comments

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+# Script for setting up OS specific CI test environment on travis-ci.org
+#
+# Copyright (C) 2016 Andreas Bok Andersen <andreas.bok@sociomantic.com>
```
daniel-zullo: 2017

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
daniel-zullo, line 1: what about `#!/bin/sh`?

travis_configure.sh
andreas-bok-sociomantic: True, fixed

daniel-zullo (commented):
I haven't finished my reviewed yet

travis_configure.sh
andreas-bok-sociomantic: Sorry made a comment but somehow didn't post it. I tried with `sh` but some of the syntax is  not compatible. Furthermore `travis-ci` uses bash scripts in their examples in their github repo. So I will stick with `bash`

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
...
+# Script which is run before installation script is called
+#
+# Installs fpm needed for deb packaging
+#
+# Returns 0
+before_install()
+{
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            gem install fpm
```
daniel-zullo, line 35: I wonder if this command needs `sudo` then.

daniel-zullo (commented):
I only have a question and LGTM.

@mathias-lang-sociomantic any further comments?

travis_configure.sh
andreas-bok-sociomantic: It doesn't. But I will try to add a printout of the gem env variables to see where gems are installed into. Might install it into $HOME/.gems which wouldn't require sudo

andreas-bok-sociomantic: Ok checked the docs for travis. They don't use sudo for their examples of installing gems, 
https://docs.travis-ci.com/user/languages/ruby/

daniel-zullo: fair enough

andreas-bok-sociomantic:
ping @mathias-lang-sociomantic

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+# Script for setting up OS specific CI test environment on travis-ci.org
+#
+# Copyright (C) 2017 Andreas Bok Andersen <andreas.bok@sociomantic.com>
```
mathias-lang-sociomantic: Sociomantic labs GmbH

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
...
+            # Ensures MySQL is started with binlog enabled. This is done by
+            # modifying the default my.cnf file.
+            #  1. Enable replication in mysql config file
+            #  2. Restart MySQL server
+            enable_mysqlbinlog "/etc/mysql/my.cnf"
+            sudo service mysql restart
+            mysql -u root -e "SET PASSWORD FOR 'root'@'localhost' = PASSWORD('')"
+        elif [[ "$DIST_PACKAGE_TARGET" == "RPM" ]]; then
+            # build a docker image with Centos 7 and MySQL
+            docker build -t centos7_mysql56 -f docker/Dockerfile .
```
mathias-lang-sociomantic, line 73: else fail ?

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
...
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            autoreconf -fi
+            ./configure
+            make check
+            make deb
+        elif [[ "$DIST_PACKAGE_TARGET" == "RPM" ]]; then
+            docker run -d --name db_container centos7_mysql56
+            docker exec -u root db_container sh -c \
+                "autoreconf -fi && ./configure && make rpm"
```
mathias-lang-sociomantic, line 102: Ditto else fail

travis_configure.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+# Installs fpm needed for deb packaging
+#
+# Returns 0
+before_install()
+{
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
+            gem install fpm
+        fi
```
mathias-lang-sociomantic, line 36: fpm is here by default on rpm or not used ?

andreas-bok-sociomantic: **fpm** is only used to build the deb package. The **rpm** package is built on the docker image. 
Just to avoid installing unnecessary apps.

travis_configure.sh
andreas-bok-sociomantic: Hmm, yes. I gues it could be good to add a notice about an invalid build configuration

andreas-bok-sociomantic: But it is not in all cases where we wan't to exit with an error. the `before_install` script should only handle the linux/deb case

mathias-lang-sociomantic: In this case make it explicit with a comment. The underlying concern was that, if you want to add an architecture / test, you don't want to miss out on the configuration somewhere (although it will most likely be visible if you do :) )

andreas-bok-sociomantic: Sure, the build matrix does expand exponentially when whenever a new build setting is added. So handling invalid cases would be helpful

andreas-bok-sociomantic:
Addressed comments by @mathias-lang-sociomantic

mathias-lang-sociomantic (approved):
LGTM

PR #175, opened 2017-05-28
Fix invalid parsing of MYSQL_PORT env variable

andreas-bok-sociomantic:
Fixes a bug where the env variable MYSQL_PORT used
as connection parameter in the unittests was checked
for but not read correctly.


mathias-lang-sociomantic:
How did the tests pass before ?

andreas-bok-sociomantic:
> How did the tests pass before ?

First reason is that MYSQL_PORT wasn't exported as env variable, so DRIZZLE_DEFAULT_TCP_PORT was always used. 
But passing  port as 0, also works both the unittests and the mysql cli client. However now it works correctly

PR #178, opened 2017-05-30
Bump version to v5.6.3

andreas-bok-sociomantic:
New ABI version is 10.2.3


mathias-lang-sociomantic:
The merge should be separate from the P.R.

andreas-bok-sociomantic:
Fixed

PR #185, opened 2017-06-13
Use beaver to upload Debian packages to bintray

leandro-lucarella-sociomantic:
This repo is quite different from other Sociomantic repos, so for now
I'm just using beaver to upload the packages.

I wonder if we should just keep this repo independent from other
Sociomantic infrastructure and copy the beaver script to upload files
via jfrog here (or just call jfrog in a custom script).

In any case, we need to have packages available via an apt repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,11 @@ env:
 
 before_install:
   - gem install fpm
+  - if [[ -n "$TRAVIS_TAG" ]]; then
+        curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
+        chmod a+x /tmp/jfrog ;
+        sudo cp /tmp/jfrog /usr/local/bin/jfrog ;
+      fi
   - mysql -u root -e "SET PASSWORD FOR 'root'@'localhost' = PASSWORD('')"
 
 ## This script runs before the compiling
@@ -63,4 +68,15 @@ script:
   - make
   - make check
   - make deb
+
+# Deploy debian packages to bintray
+deploy:
+    provider: script
+    script: beaver bintray upload -N pkg/deb/*.deb
+    skip_cleanup: true
+    on:
+        tags: true
+        condition: >
+            "$TRAVIS_OS_NAME" = "linux" -a
+            "$CC" = "gcc"
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "beaver"]
+	path = beaver
+	url = https://github.com/sociomantic-tsunami/beaver.git


mathias-lang-sociomantic:
Travis says no :(

leandro-lucarella-sociomantic:
Yeah, I don't understand why it won't accept the `on:` conditions, according to my interpretation of [the (poor) docs](https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A) it should work.

I get:
```
$ travis lint 
Warnings for .travis.yml:
[x] syntax error: (<unknown>): did not find expected key while parsing a block mapping at line 103 column 9
```

leandro-lucarella-sociomantic:
OK, it was the `"` character in YML. Is funny, I could never find a good user-oriented YML document, and the specs are a bit too complicated to use as user documentation.

travis_configure.sh
```diff
@@ -35,6 +35,11 @@ before_install()
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
         if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
             gem install fpm
+            if [[ -n "$TRAVIS_TAG" ]]; then
```
andreas-bok-sociomantic: Could you add `jprog` to the dependency which is installed in the doc. Something like

```
# Installs dependencies
# linux:
#   deb:
#     - fpm
#     - jfrog (If the current build is for a git tag)
```

andreas-bok-sociomantic (commented):
LGTM, just one comment

andreas-bok-sociomantic:
With reference to #185, should we include this in the patch release as well, @leandro-lucarella-sociomantic

travis_configure.sh
leandro-lucarella-sociomantic: Well, jfrog is not needed to install anything, it's just needed to automatically upload files to bintray. If you build locally, even if you build the package, you don't need `jfrog`.

leandro-lucarella-sociomantic:
Maybe we could include it. It might fail, but we could do another patch release if needed before going with the full train.

leandro-lucarella-sociomantic:
Rebased. Travis support changed quite a bit between v5.5.x and v5.x.x :unamused:

travis_configure.sh
andreas-bok-sociomantic: I see. In that it is fine as is

andreas-bok-sociomantic:
> Rebased. Travis support changed quite a bit between v5.5.x and v5.x.x :unamused:

Yeah sorry, should have mentioned that.

.travis.yml
```diff
@@ -64,3 +69,15 @@ script:
...
+# Deploy debian packages to bintray
+deploy:
+    provider: script
+    script: beaver bintray upload -N pkg/deb/*.deb
+    skip_cleanup: true
+    on:
+        tags: true
+        condition: >
+            "$TRAVIS_OS_NAME" = "linux" -a
+            "$DIST_PACKAGE_TARGET" = "DEB" -a
```
andreas-bok-sociomantic: 1. I would assume that **DIST_PACKAGE_TARGET** should be declared under the `env` to be used in the comparison. 
2. With the current .travis settings, the conditions `$TRAVIS_OS_NAME == linux` and `"$CC" = "gcc"` would be sufficient to distinguish the build. I.e **DIST_PACKAGE_TARGET** isn't needed until the changes added in `v5.x.x`

leandro-lucarella-sociomantic: I removed it, it was a leftover from v5.x.x, good catch!

andreas-bok-sociomantic:
Just a question. Can you confirm that the deb is uploaded to the bintray? I checked the travis build log and can't see that it enters the deploy step and executes the script

andreas-bok-sociomantic:
> Can you confirm that the deb is uploaded to the bintray? 

Of course it doesn't perform the step since the build isn't for a git tag. Maybe try with a dummy tag just to see if it works

leandro-lucarella-sociomantic:
> Of course it doesn't perform the step since the build isn't for a git tag. Maybe try with a dummy tag just to see if it works

The script worked for a few other repos already, what it could fail is more how is it called from travis, because is the only part that changes from project to project. I would say let's just merge it and tag and see what happens. If something goes wrong, we fix it and tag a new version. If you want to be really conservative, just tag a pre-release, like v5.5.5-rc1 and if it works you can tag the same commit as the final release.

andreas-bok-sociomantic:
> let's just merge it and tag and see what happens

Right on. I will try to tag the **bump version PR** and see what happens. 
Patch release under way :)

andreas-bok-sociomantic:
LGTM

PR #188, opened 2017-06-16
Bug: fix travis deploy script 

andreas-bok-sociomantic:
In spite of no build error the deploy step 
didn't succeed. 

Travis wraps the condition inside double brackets.
Thus && is expected as comparison operator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,13 @@ services:
 #  Please refer to the section 'script' for info on
 #  SKIP_TEST_ON_ERROR
 env:
-  - MYSQL_SERVER=127.0.0.1 \
-    MYSQL_PASSWORD='' \
-    MYSQL_USER=root \
-    SKIP_TEST_ON_ERROR=0
+  global:
+    - MYSQL_SERVER=127.0.0.1
+    - MYSQL_PASSWORD=''
+    - MYSQL_USER=root
+    - SKIP_TEST_ON_ERROR=0
+    - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
+
 
 ## Defines packages to be installed by apt
 #  2016/12/01 - mysql-5.6 was added to the default stack
@@ -69,14 +72,11 @@ script:
   - make check
   - make deb
 
-# Deploy debian packages to bintray
+# Deploy debian packages to bintray when building for a git tag
 deploy:
     provider: script
-    script: beaver bintray upload -N pkg/deb/*.deb
+    script: beaver bintray upload -N deb/*.deb
     skip_cleanup: true
     on:
         tags: true
-        condition: >
-            "$TRAVIS_OS_NAME" = "linux" -a
-            "$CC" = "gcc"
-
+        condition: $TRAVIS_OS_NAME = 'linux' && $CC = 'gcc'


.travis.yml
```diff
@@ -72,11 +72,8 @@ script:
 # Deploy debian packages to bintray
 deploy:
     provider: script
-    script: beaver bintray upload -N pkg/deb/*.deb
+    script: beaver bintray upload -N deb/*.deb
```
andreas-bok-sociomantic: Should the `upload` script be called with the full path? 
Travis errors with this for the deploy step:  

```
Preparing deploy

Deploying application
sh: 1: beaver: not found

Script failed with status 127
failed to deploy
```

leandro-lucarella-sociomantic: Ahhh, right! That would be one choice, but it's better to add it to the path:

```ỳml
env:
    global:
        - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
```

leandro-lucarella-sociomantic: BTW, are you sure this should be `deb/*.deb` and not `pkg/deb/*.deb`?

.travis.yml
```diff
@@ -72,11 +72,8 @@ script:
...
     skip_cleanup: true
     on:
         tags: true
-        condition: >
-            "$TRAVIS_OS_NAME" = "linux" -a
-            "$CC" = "gcc"
-
+        condition: $TRAVIS_OS_NAME = 'linux' && $CC = 'gcc'
```
leandro-lucarella-sociomantic, line 82: And do you know this is really a problem? Because it should work. I mean, putting it in one line seems good if it fits but the `-a` -> `&&` shouldn't be necessary if the condition is supposed to be pasted in at `if [[ ... ]]`.

.travis.yml
andreas-bok-sociomantic: Yeah, the commit, https://github.com/sociomantic-tsunami/libdrizzle-redux/commit/b708fd9e263d559332a9fda736f22fe6c6cd3f4c,  which moves the packing scripts to the `pkg` is on the 5.x.x. MIssed that when reviewing after the rebase on v5.5.x

.travis.yml
andreas-bok-sociomantic: Ah. So is it quoting of the conditions which maybe messes up the if statement? 
The parsing error for the `-a` syntax is : 

```
[0K/home/travis/.travis/job_stages: line 558: conditional binary operator expected
/home/travis/.travis/job_stages: line 558: expected `)'
/home/travis/.travis/job_stages: line 558: syntax error near `-a'
/home/travis/.travis/job_stages: line 558: `  if [[ ("$TRAVIS_OS_NAME = linux" -a "$CC = gcc"'
/home/travis/.travis/job_stages: line 558: conditional binary operator expected
/home/travis/.travis/job_stages: line 558: expected `)'
```

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic 
`deploy` step is reached but fails with invalid cmd args since **BINTRAY_USER** is not defined. 
Does the values for **$BINTRAY_USER** and **$BINTRAY_KEY** correspond to github user and pass?

leandro-lucarella-sociomantic:
No, those are the credentials to upload to the sociomantic-tsunami bintray repos. If you reached that point, we need to continue the testing in the blessed GitHub repo and the real bintray repos. At least unless you want to create your own bintray user and debian repo and upload there. But IMHO is not worth the trouble.

.travis.yml
leandro-lucarella-sociomantic: Really? that's messed up. I sometimes don't understand what is Travis trying to do...

andreas-bok-sociomantic: It is quite strange yes. The file with `-a` syntax passes the travis' online linter. 
I found the fix by trial and error. Would setting verbose flag for the shell give more insight into the mysterious ways of travis?

andreas-bok-sociomantic:
> No, those are the credentials to upload to the sociomantic-tsunami bintray repos. 

Oki 
> If you reached that point, we need to continue the testing in the blessed GitHub repo and the real bintray repos

Once the syntax for the `deploy.condition` is settled I will do the patch release. And then the actual testing of the upload will then happen post v5.5.5

leandro-lucarella-sociomantic:
> Once the syntax for the deploy.condition is settled I will do the patch release. And then the actual testing of the upload will then happen post v5.5.5

The upload of the packages will happen as soon as you tag v5.5.5 and Travis tests for that tag pass :)

I suggest testing the package upload by making a v5.5.5-rc1 tag, this will upload the package as a pre-release, letting know the users this is not a final release (in the GitHub releases page you can also mark it as a pre-release).

andreas-bok-sociomantic:
Wrt to the `deploy.on.condition`. Unless you insist I don't feel it is worth spending more time debugging the 
travis'

andreas-bok-sociomantic:
Squashed commits

.travis.yml
leandro-lucarella-sociomantic: Ahh, ok, sorry.

.travis.yml
leandro-lucarella-sociomantic: Yeah, usually adding `set -uxe` is a good idea, but with deployment scripts is also strange that you can only provide a one-liner, so you have to separate all commands via `;`.

leandro-lucarella-sociomantic:
> configtion

Condition I presume? Yes, move forward with it, I was just curious!

.travis.yml
andreas-bok-sociomantic: Tried setting the the options with `bash -c "set -eux"` 
However  AFAICS it doesn't affect  the verbosity output compared to earlier builds `.travis`.
However we might set the options in v5.x.x which adds a separate bash script. 
Did experience though that the build errored due to the extensive logging after trying `set -v`

andreas-bok-sociomantic:
Unblocked

.travis.yml
```diff
@@ -22,10 +22,13 @@ services:
...
 env:
-  - MYSQL_SERVER=127.0.0.1 \
-    MYSQL_PASSWORD='' \
-    MYSQL_USER=root \
-    SKIP_TEST_ON_ERROR=0
+  global:
+    - MYSQL_SERVER=127.0.0.1 \
+      MYSQL_PASSWORD='' \
+      MYSQL_USER=root \
+      SKIP_TEST_ON_ERROR=0
```
leandro-lucarella-sociomantic: Why not using one item per variable now? That format looks a bit strange.

leandro-lucarella-sociomantic (approved):
Other than the minor comment, LGTM

.travis.yml
andreas-bok-sociomantic: Sure

andreas-bok-sociomantic:
Updated based on comments from @leandro-lucarella-sociomantic

.travis.yml
leandro-lucarella-sociomantic: no, without the bash, just:
```yml
script: set -eux; beaver bintray ...
```

andreas-bok-sociomantic:
Ah

leandro-lucarella-sociomantic:
It still LGTM. Feel free to auto-merge unless you expect someone else to review this.

andreas-bok-sociomantic:
Oki, I will just test the `set -eux` before merging

andreas-bok-sociomantic:
> Oki, I will just test the set -eux before merging

Didn't see any change on the build log. I will merge it without the `set -eux`, but keep it in mind for the changes in `v5.x.x`. Tx for the reviews

andreas-bok-sociomantic:
Just a question, should it be released release candidate, ie. `v5.5.5-rc1`?

leandro-lucarella-sociomantic:
I would do that, yes. If all goes well we can tag the same commit as final, if not we add more commits to fix the problems before the final v5.5.5 release (and then merge to the other branches and release those too).

leandro-lucarella-sociomantic:
> Didn't see any change on the build log. I will merge it without the set -eux, but keep it in mind for the changes in v5.x.x. Tx for the reviews

The change should be there but minimal, only a `+beaver bintray ...` should be added to the output.

andreas-bok-sociomantic:
> The change should be there but minimal,

Ah ok

andreas-bok-sociomantic:
> . If all goes well we can tag the same commit as final, if not we add more commits to fix the problems before the final v5.5.5 release 

All went well. So I will do the patch release with the tag `v5.5.5`, right? OAIMS

PR #196, opened 2017-07-14
Bugfix: Use autoupdate before install for Homebrew

andreas-bok-sociomantic:
Without update before install some recipes are
no longer valid. E.g. the download url from bintray
has changed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ env:
     - MYSQL_PORT=3306
     - DIST_PACKAGE_TARGET=DEB
     - SKIP_TEST_ON_ERROR=0
-    - HOMEBREW_NO_AUTO_UPDATE=1
     - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
   matrix:
     - DIST_PACKAGE_TARGET=DEB


.travis.yml
```diff
@@ -47,7 +47,7 @@ env:
     - MYSQL_PORT=3306
     - DIST_PACKAGE_TARGET=DEB
     - SKIP_TEST_ON_ERROR=0
-    - HOMEBREW_NO_AUTO_UPDATE=1
+    - HOMEBREW_NO_AUTO_UPDATE=0
```
gautam-kotian-sociomantic: Negatively named boolean variables take so much longer to understand. This would've been much simpler if it was:
```
- HOMEBREW_AUTO_UPDATE=1
```
But anyway, that's not really a concern of this PR.

gautam-kotian-sociomantic (approved):
LGTM

.travis.yml
andreas-bok-sociomantic: Haha, true. I could make a PR in the travis repo :)

mathias-lang-sociomantic: I would assume auto update is the default then ?

andreas-bok-sociomantic: Ah yeah, actually could just remove the setting. Will do. As @gautam-kotian-sociomantic said double negations makes you loose track

andreas-bok-sociomantic:
Updated 
- Rebased on latest v5.x.x branch
- Remove setting instead of explicitly turning it ON

mathias-lang-sociomantic (approved):

ben-palmer-sociomantic:
LGTM

PR #152, opened 2017-04-20
Add rst documentation for deprecated enums

andreas-bok-sociomantic:
- Add missing rst documentation for drizzle_socket_option_t
- Replace drizzle_socket_owner with drizzle_socket_owner_t

PR #200, opened 2017-08-11
Fix crash when SQL gives a single NULL

chenzhuoyu:
`libdrizzle-redux` would trigger a crash when SQL statements gives only `NULL` fields, this is caused by not handling fields properly in function `drizzle_field_read`.

We can test this by a simple SQL statement:
```sql
SELECT NULL
```

In `drizzle_field_read`(`libdrizzle/field.cc`):
```cpp
/* line 69 ~ 92 */
  if (result->has_state())
  {
    if (result->field_current == (result->column_count - result->null_bitcount))
    {
      *ret_ptr= DRIZZLE_RETURN_ROW_END;
      return NULL;
    }

    if (result->binary_rows)
    {
      result->push_state(drizzle_state_binary_field_read);
    }
    else
    {
      result->push_state(drizzle_state_field_read);
    }
  }

  if (result->binary_rows && (result->field_current == 0))
  {
    result->push_state(drizzle_state_binary_null_read);
  }

  *ret_ptr= drizzle_state_loop(result->con);
```

Here we can see, when the first time `drizzle_field_read` been called, it pushes 2 states into the state machine, which are `drizzle_state_binary_null_read` for handling `NULL` fields, and `drizzle_state[_binary]_field_read` for handling non-`NULL` fields.

If the SQL statement gives only `NULL` fields, the `drizzle_state[_binary]_field_read` state should **NOT** be pushed into state machine, since there are no actual fields to be read.

This patch fixes this problem by forcing a `drizzle_state_binary_null_read` before handling non-`NULL` fields, this enables the `ROW_END` checking before pushing `drizzle_state[_binary]_field_read` into state machine, thus solves the problem.

I will construct a simple test program to help solving this problem.

### DIFF
--- a/tests/unit/nulls.c
+++ b/tests/unit/nulls.c
@@ -146,6 +146,23 @@ int main(int argc, char *argv[])
   CHECKED_QUERY(querybuf);
 
 #ifdef TEST_PREPARED_STATEMENTS
+  strcpy(querybuf, "SELECT NULL");
+  sth = drizzle_stmt_prepare(con, querybuf, strlen(querybuf), &driz_ret);
+
+  ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK, "Error (%s): %s, preparing \"%s\"",
+             drizzle_strerror(driz_ret), drizzle_error(con), querybuf);
+
+  driz_ret = drizzle_stmt_execute(sth);
+  ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK, "Error (%s): %s, executing \"%s\"",
+             drizzle_strerror(driz_ret), drizzle_error(con), querybuf);
+
+  driz_ret = drizzle_stmt_buffer(sth);
+  ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK,
+             "Error (%s): %s, buffering result from query \"%s\"",
+             drizzle_strerror(driz_ret), drizzle_error(con), querybuf);
+
+  ASSERT_EQ(drizzle_stmt_close(sth), DRIZZLE_RETURN_OK);
+
   strcpy(querybuf, "INSERT INTO test_nulls.t1 VALUES (?,?,?,?,?,?,?,?,?,?,?)");
   sth = drizzle_stmt_prepare(con, querybuf, strlen(querybuf), &driz_ret);
   ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK, "Error (%s): %s, preparing \"%s\"",
--- a/libdrizzle/field.cc
+++ b/libdrizzle/field.cc
@@ -66,6 +66,16 @@ drizzle_field_t drizzle_field_read(drizzle_result_st *result, uint64_t *offset,
     return 0;
   }
 
+  if (result->binary_rows && (result->field_current == 0))
+  {
+    result->push_state(drizzle_state_binary_null_read);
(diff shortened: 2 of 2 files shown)


chenzhuoyu:
Here is the test program on my Gist: https://gist.github.com/chenzhuoyu/7df3ae43131dc658d2659f7c3783fab3

Hope it helps :)

chenzhuoyu:
There are 2 build jobs failed with the same error:
```
Invalid package configuration: Cannot package the path './install/usr/lib/libdrizzle-redux.so', does it exist? {:level=>:error}
```

But other build jobs are fine, any information ? @andreas-bok-sociomantic 
I can build the library on my computer without any problem ...

andreas-bok-sociomantic:
Thanks @chenzhuoyu. The ci build error is caused by a bug in v1.9.x releases of `fpm`, 
cf. https://github.com/jordansissel/fpm/issues/1395, and bugfix https://github.com/jordansissel/fpm/pull/1399

On holiday atm, but will take a look first thing Monday.

andreas-bok-sociomantic:
Reproduced the error locally. 
Added your gist to the `nulls` unittest on this branch, https://github.com/andreas-bok-sociomantic/libdrizzle-redux/tree/sql-null-crash-fix-testcase. You can just cherrypick that commit and add to this PR. 
Also changed the base branch to v5.5.x as I want the fix to go into the next patch release v5.5.6. 
Thanks again :)

chenzhuoyu:
@andreas-bok-sociomantic Test case added ;)

andreas-bok-sociomantic:
Super, could you rebase on the latest `v5.5.x` branch. It fixes the travis build error.

andreas-bok-sociomantic:
LGTM, after rebase

chenzhuoyu:
@andreas-bok-sociomantic It builds successfully now ;)

tests/unit/nulls.c
```diff
@@ -146,6 +146,23 @@ int main(int argc, char *argv[])
...
 #ifdef TEST_PREPARED_STATEMENTS
+  strcpy(querybuf, "SELECT NULL");
+  sth = drizzle_stmt_prepare(con, querybuf, strlen(querybuf), &driz_ret);
+
+  ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK, "Error (%s): %s, preparing \"%s\"",
+             drizzle_strerror(driz_ret), drizzle_error(con), querybuf);
+
+  driz_ret = drizzle_stmt_execute(sth);
+  ASSERT_EQ_(driz_ret, DRIZZLE_RETURN_OK, "Error (%s): %s, executing \"%s\"",
+    drizzle_strerror(driz_ret), drizzle_error(con), querybuf);
```
gautam-kotian-sociomantic: Why the different indentation on this line (as against in the `ASSERT_EQ_()` statements before and after)?

andreas-bok-sociomantic: Should be the same as in line [\#152](https://github.com/sociomantic-tsunami/libdrizzle-redux/pull/200/files/a90efb587f9c1ce5af55af7637278b2cfaddc95b#diff-02a60f8dc8105e273191327adaaf11b6R152)
@chenzhuoyu could you fix that?

chenzhuoyu:
@andreas-bok-sociomantic Fixed

andreas-bok-sociomantic:
Great, I will merge, thanks :)

PR #202, opened 2017-08-12
Check for NULL params in drizzle_binlog_init function

andreas-bok-sociomantic:
It wouldn't make sense to use the binlog without the callbacks. 
But no check is made as to whether the  event and error callback functions are not **NULL** before they are invoked. 

Add checks to the binlog unittest for passing NULL arguments to drizzle_binlog_init


ben-palmer-sociomantic:
LGTM

PR #206, opened 2017-08-16
Bugfix: Set env variable MYSQL_PASSWORD=pass in dockerfile

andreas-bok-sociomantic:
The travis build failed for RPM which builds the library in a centos docker container

The MYSQL_PASSWORD was previously set to an empty string as '' which was not allowed.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
 # Set environment variables
 ENV MYSQL_USER=user \
     MYSQL_PASSWORD=pass \
-    MYSQL_ROOT_PASSWORD='' \
+    MYSQL_ROOT_PASSWORD=pass \
     MYSQL_SERVER=127.0.0.1 \
     MYSQL_PORT=3306 \
     MYSQL_DATABASE=test \


docker/Dockerfile
```diff
@@ -7,7 +7,7 @@ MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
 # Set environment variables
 ENV MYSQL_USER=user \
     MYSQL_PASSWORD=pass \
-    MYSQL_ROOT_PASSWORD='' \
+    MYSQL_ROOT_PASSWORD=pass \
```
gautam-kotian-sociomantic, line 10: Why "pass"? If this is supposed to be a dummy password, wouldn't it be better to make it more self-documenting as `MYSQL_ROOT_PASSWORD=dummy_password` or something like that?

andreas-bok-sociomantic: I see what you mean. In the unittest it is the **MYSQL_USER** which is used, so setting **MYSQL_ROOT_PASSWORD** is only to comply with the policy of MySQL that root should have a password. 
I would rather keep as is, if it's ok with you. It is only exposed to the Travis CI, so the user of library wouldn't have to deal with this env variable anyway

gautam-kotian-sociomantic: Yeah sure, fine by mean if you prefer to keep it as-is.

andreas-bok-sociomantic: I would :)

PR #207, opened 2017-08-16
Make binlog unittest compile with GCC

andreas-bok-sociomantic:
Fixes issue where binlog unittest was compiled as C++ possibly breaking the C API.

- Remove C++ imports
- Rename the file to binlog.c
- Declare prototypes for callback functions


ben-palmer-sociomantic:
LGTM

PR #213, opened 2017-08-18
Bugfix: invalid parsing of binlog event data

andreas-bok-sociomantic:
The fixed bytesize binlog header was incorrectly
defined as 27.

### DIFF
--- a/libdrizzle/binlog.cc
+++ b/libdrizzle/binlog.cc
@@ -324,9 +324,12 @@ drizzle_return_t drizzle_state_binlog_read(drizzle_st *con)
     con->buffer_size--;
     binlog_event->raw_data= con->buffer_ptr;
     binlog_event->timestamp= drizzle_get_byte4(con->buffer_ptr);
-    binlog_event->type= (drizzle_binlog_event_types_t)con->buffer_ptr[4];
-    binlog_event->server_id= drizzle_get_byte4(con->buffer_ptr + 5);
-    binlog_event->raw_length= binlog_event->length= drizzle_get_byte4(con->buffer_ptr + 9);
+    binlog_event->type=
+      (drizzle_binlog_event_types_t)con->buffer_ptr[DRIZZLE_EVENT_POSITION_TYPE];
+    binlog_event->server_id= drizzle_get_byte4(
+      con->buffer_ptr + DRIZZLE_EVENT_POSITION_SERVERID);
+    binlog_event->raw_length= binlog_event->length= drizzle_get_byte4(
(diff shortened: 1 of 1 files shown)


ben-palmer-sociomantic:
Should the comparison here https://github.com/andreas-bok-sociomantic/libdrizzle-redux/blob/feeb4f7b1405f05c99e604c61e47c3049a28b0b6/libdrizzle/binlog.cc#L337 also be adjusted?

andreas-bok-sociomantic:
> Should the comparison here https://github.com/andreas-bok-sociomantic/libdrizzle-redux/blob/feeb4f7b1405f05c99e604c61e47c3049a28b0b6/libdrizzle/binlog.cc#L337 also be adjusted?

Looked at the MySQL binlog documentation again. AFAICS no event will ever have a length <=27 for all version of the binlog, https://dev.mysql.com/doc/internals/en/binary-log-versions.html 

I tried adding an assert(false) in the first line of the section. Binlog unittest still succeeds. Thus quite sure the code will never be reached and can be safely removed.

ben-palmer-sociomantic:
> I tried adding an assert(false) in the first line of the section. Binlog unittest still succeeds. Thus quite sure the code will never be reached and can be safely removed.

Ok maybe just change it to an assert that the length is at least the size of the header or whatever else is the minimum required (19?).

Is this related to https://github.com/sociomantic-tsunami/libdrizzle-redux/issues/214 ?

andreas-bok-sociomantic:
> Ok maybe just change it to an assert that the length is at least the size of the header or whatever else is the minimum required (19?)

Yes, I would also prefer to keep the code and add an assert

> Is this related to #214 ?

No

andreas-bok-sociomantic:
Looked at the documentation again. It is actually the case that event types **APPEND_BLOCK_EVENT, EXEC_LOAD_EVENT, DELETE_FILE_EVENT** have length of 27. The layout is 
```
Fixed Header                  19
Fixed data part                4
Variable data part             0
Checksum                       4

```
https://dev.mysql.com/doc/internals/en/load-data-infile-events.html

I would like to look further into this issue before adding an `assert(false)`.

ben-palmer-sociomantic:
> I would like to look further into this issue before adding an assert(false).

Cool I blocked it just to be clear.

andreas-bok-sociomantic:
> Cool I blocked it just to be clear.

Cool, the current changes though are good to go, so I will unblock.

ben-palmer-sociomantic:
LGTM.

Issue #214, opened 2017-08-19
Off-by-one error in drizzle_check_unpack_error()?


kengruven:
When I run MySQL's command-line client with a fake schema name, it reports the error like this:

    $ mysql -uroot badname
    ERROR 1049 (42000): Unknown database 'badname'

When I try to connect using libdrizzle-redux:

        conn = drizzle_create("localhost", 3306, "root", "", "badname", nil)
        ok = drizzle_connect(conn)

the C string I get from `drizzle_error(conn)` is:

    "drizzle_check_unpack_error: Unknown database 'badname"

There's no ending `'` here.  When I stop the debugger inside drizzle_check_unpack_error(), I can see:

    con->error_code = (uint16_t) 1049
    con->sqlstate = (char [6]) "42000"
    con->buffer_ptr = (unsigned char *) "Unknown database 'badname't"
    con->last_error = (char [2048]) "Unknown database 'badname"

It looks like it's receiving the full message, but the snprintf call is stopping one byte too early.  There is a fixed size limit of DRIZZLE_MAX_SQLSTATE_SIZE (2048) but this message is well below that (26).

I see the same issue (final character missing) for other error messages, too.

andreas-bok-sociomantic:
Reproduced the error and made a PR for a bugfix, #215
Thanks for reporting the bug :)

kengruven:
Great -- that was quick!  Thanks.

A couple remarks on error handling (which aren't bugs, really):

 - I think it would be handy to have just the server error available as a string, somehow, without the "drizzle_" function name.  It's nicer to present to users than raw MySQL error codes, and drizzle function names aren't useful to (application) users.  (I'm not sure why the drizzle function name is useful to me, even as a developer.  I know what call it was: the one I just made.)

 - There's 2 places that hardcode a function name to drizzle_set_error, rather than using `__func__`: libdrizzle/result.cc:363 and libdrizzle/row.cc:67.

PR #215, opened 2017-08-21
Bugfix: wrong parsing of server error packet

andreas-bok-sociomantic:
`drizzle_check_unpack_error` is missing one char when copying the msg in a server error packet
to `con->last_error`

Addresses #214

### DIFF
--- a/libdrizzle/pack.cc
+++ b/libdrizzle/pack.cc
@@ -496,7 +496,7 @@ bool drizzle_check_unpack_error(drizzle_st *con)
   con->packet_size-= 9;
 
   snprintf(con->last_error, DRIZZLE_MAX_ERROR_SIZE, "%.*s",
-           (int)con->packet_size-1, con->buffer_ptr);
+           (int)con->packet_size, con->buffer_ptr);
 
   drizzle_set_error(con, __func__, " %s", con->last_error);
 


mathias-lang-sociomantic:
Tests?

andreas-bok-sociomantic:
They are coming, but made quite a few changes to integrate it in the `connect` unittest. 
Another PR is underway

don-clugston-sociomantic:
Was the "-1" originally there because of a null terminator, or something?
Is this code correct when the string has length DRIZZLE_MAX_ERROR_SIZE?

andreas-bok-sociomantic:
> Was the "-1" originally there because of a null terminator, or something?

Could be the case. Probably the assumption that the packet data includes the `'\0'` terminator. Moreover `snprintf` terminates its output with `'\0'`. 

> Is this code correct when the string has length DRIZZLE_MAX_ERROR_SIZE?

Good catch. `DRIZZLE_MAX_ERROR_SIZE = 2048`, so hopefully MySQL will never emit an error string with that many characters. However the `last_error` should be declared with length `DRIZZLE_MAX_ERROR_SIZE+1`

don-clugston-sociomantic:
OK, LGTM.

don-clugston-sociomantic:
Well, looking at the definition of snprintf, it only copies n-1 bytes, and then adds a null terminator. So I think your extra change is not required. I have led you astray :).

andreas-bok-sociomantic:
Ah. Will change back :+1:

andreas-bok-sociomantic:
Reverted the change to the length of `last_error`

andreas-bok-sociomantic:
Will self merge unless you have more comments @don-clugston-sociomantic

ben-palmer-sociomantic:
> Will self merge

Too slow :)

PR #216, opened 2017-08-21
Add unittests for invalid connection parameters

andreas-bok-sociomantic:
Test error string and drizzle_return_t value
for invalid host, port, user, pass, db settings

Addresses bugfix #215

### DIFF
--- a/tests/unit/connect.c
+++ b/tests/unit/connect.c
@@ -42,6 +42,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+drizzle_st *con;
+drizzle_return_t driz_ret;
+
 int main(int argc, char *argv[])
 {
   (void)argc;
@@ -50,12 +53,12 @@ int main(int argc, char *argv[])
   drizzle_options_st *opts = drizzle_options_create();
   drizzle_socket_set_options(opts, 10, 5, 3, 3);
 
-  drizzle_st *con= drizzle_create(getenv("MYSQL_SERVER"),
-                                  getenv("MYSQL_PORT") ? atoi(getenv("MYSQL_PORT"))
-                                                       : DRIZZLE_DEFAULT_TCP_PORT,
-                                  getenv("MYSQL_USER"),
-                                  getenv("MYSQL_PASSWORD"),
-                                  getenv("MYSQL_SCHEMA"), opts);
+  con = drizzle_create(getenv("MYSQL_SERVER"),
+                       getenv("MYSQL_PORT") ? atoi(getenv("MYSQL_PORT"))
+                                            : DRIZZLE_DEFAULT_TCP_PORT,
+                       getenv("MYSQL_USER"),
+                       getenv("MYSQL_PASSWORD"),
+                       getenv("MYSQL_SCHEMA"), opts);
   ASSERT_NOT_NULL_(con, "Drizzle connection object creation error");
 
   int opt_val = drizzle_socket_get_option(con, DRIZZLE_SOCKET_OPTION_TIMEOUT);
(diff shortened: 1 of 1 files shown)


andreas-bok-sociomantic:
Blocked on #215

andreas-bok-sociomantic:
Updated and unblocked 
- Rebased on latest v5.5.x

tests/unit/connect.c
```diff
@@ -92,6 +92,35 @@ int main(int argc, char *argv[])
...
+  test_connection_error(host, 1234, user, pass, db,
+    DRIZZLE_RETURN_COULD_NOT_CONNECT, "drizzle_state_connect:could not connect",
+    -1);
+
+  // invalid user
+  sprintf(error_expected, "drizzle_check_unpack_error: Access denied for user "
+    "'invalid_user'@'localhost' (using password: %s)",
+    (pass[0] == '\0') ? "NO" : "YES");
+  test_connection_error("localhost", port, "invalid_user", pass, db,
+  DRIZZLE_RETURN_HANDSHAKE_FAILED, error_expected, -1);
```
ben-palmer-sociomantic: Inconsistent indentation here.

tests/unit/connect.c
```diff
@@ -92,6 +92,35 @@ int main(int argc, char *argv[])
...
+  test_connection_error("1.2.3.4", port, user, pass, db, DRIZZLE_RETURN_TIMEOUT,
+    "drizzle_wait:timeout reached", 3);
+
+  // invalid port
+  test_connection_error(host, 1234, user, pass, db,
+    DRIZZLE_RETURN_COULD_NOT_CONNECT, "drizzle_state_connect:could not connect",
+    -1);
+
+  // invalid user
+  sprintf(error_expected, "drizzle_check_unpack_error: Access denied for user "
```
ben-palmer-sociomantic: Out of interest why are we using `sprintf` here to create the expected string rather than just hard coding it?

tests/unit/connect.c
```diff
@@ -49,16 +49,18 @@ int main(int argc, char *argv[])
 {
   (void)argc;
   (void)argv;
-
-  drizzle_options_st *opts = drizzle_options_create();
+  drizzle_options_st *opts;
```
ben-palmer-sociomantic: Could you explain in the commit message why you are making this change as its not obvious just looking at it.

ben-palmer-sociomantic (commented):
I had a couple of small comments.

tests/unit/connect.c
andreas-bok-sociomantic: The pass is read from the env varr MYSQL_PASSWORD. The end of the error message is `(using password : YES )` or `(using password : NO )` depending on the value of that env var. 
However do see that using `sprintf` can be removed if we hardcode the value for `pass` in this case. 
This could be applied to the other cases as well and thus simplify the code. Seems like a better option since the connection parameters are somewhat irrelevant for these tests. 
Thus will hardcode the parameters.

tests/unit/connect.c
andreas-bok-sociomantic: It was solely to define all variables at the top of the `main` function. But see that it is unnecessary given that the options are not used in the connection failure tests. 
Will revert this change when addressing the comments

andreas-bok-sociomantic:
Addressed comments by @ben-palmer-sociomantic 
- Hardcoded expected error messages
- Still copy expected error message for strings over 80 chars. But use `strcpy` instead of `sprintf`

ben-palmer-sociomantic:
LGTM

PR #218, opened 2017-08-21
Add OS name and version to Dockerfile filename

andreas-bok-sociomantic:
Use the operating system and version to distinguish between `Dockerfiles`


ben-palmer-sociomantic:
The commit message doesn't mention the moving of one file.

ben-palmer-sociomantic:
Ha oh I see yes it does, ignore me :)

PR #224, opened 2017-08-25
Remove test of error messages in connect unittest

andreas-bok-sociomantic:
Fixes an issue where error messages are not the same for different versions of MySQL.

Problem was that the **RPM** package is built in a docker image running `centos`. But that CI build was not part of the latest release, so everything looked good until `v5.6.6` was merged into `v5.x.x` :disappointed:

### DIFF
--- a/tests/unit/connect.c
+++ b/tests/unit/connect.c
@@ -64,18 +64,17 @@ drizzle_return_t driz_ret;
 void test_connection_error(const char *host, in_port_t port,
                            const char *user, const char *password,
                            const char *db, drizzle_return_t ret_expected,
-                           const char *error_expected, int timeout);
+                           int timeout);
 
 void test_connection_error(const char *host, in_port_t port,
                            const char *user, const char *password,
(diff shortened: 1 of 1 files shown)


ben-palmer-sociomantic:
Thats frustrating they are different in different mysql versions. Is that for all the messages?

andreas-bok-sociomantic:
> Thats frustrating they are different in different mysql versions. Is that for all the messages?

Yes, I was a bit to quick with the unittest in the initial PR. Problem is that the RPM package is built in a docker image running centos. So could in reality also be the combination of a different OS and the MySQL version. But that CI build was not part of the latest release, so everything looked good until I merged `v5.6.6` into `v5.x.x` :disappointed:

andreas-bok-sociomantic:
@ben-palmer-sociomantic if you are happy with the changes you are welcome to merge. AFAICS this is the PR which will make v5.x.x turn :green_heart:

ben-palmer-sociomantic:
LGTM

PR #225, opened 2017-08-29
Don't deploy rpm packages to bintray

andreas-bok-sociomantic:
At the moment we don't want to deploy `rpm` packages to `bintray`


ben-palmer-sociomantic:
Don't know too much about travis setup but this LGTM.

PR #226, opened 2017-08-29
travis: Fix wrong deb pkg path argument

andreas-bok-sociomantic:
On `v5.x.x` the location layout for the generated deb packages were changed. `deb/*.deb` -> `pkg/deb/*.deb`. 
Since tagged builds were created for `v5.x.x` until now, the problem with the wrong path was not discovered earlier.


ben-palmer-sociomantic:
LGTM

PR #227, opened 2017-09-02
Update and simplify packaging scripts

andreas-bok-sociomantic:
Pass common fpm args as variables 

Update deb and rpm package descriptions


ben-palmer-sociomantic:
LGTM

PR #231, opened 2017-09-12
travis: Pass env vars to docker rpm build

andreas-bok-sociomantic:
Fixes a problem where the configuration in
the centos/mysql Dockerfile is not persisted
after `docker run` is done

### DIFF
--- a/travis_configure.sh
+++ b/travis_configure.sh
@@ -129,7 +129,10 @@ run_tests()
             make deb
         elif [[ "$DIST_PACKAGE_TARGET" == "RPM" ]]; then
             docker run -d --name db_container centos7_mysql56
-            docker exec -u root db_container sh -c \
+            docker exec \
+                -e MYSQL_USER=$MYSQL_USER \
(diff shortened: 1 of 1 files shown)


travis_configure.sh
```diff
@@ -129,7 +129,8 @@ run_tests()
             make deb
         elif [[ "$DIST_PACKAGE_TARGET" == "RPM" ]]; then
             docker run -d --name db_container centos7_mysql56
-            docker exec -u root db_container sh -c \
+            docker exec -e MYSQL_USER=root -e MYSQL_PASSWORD= \
+                -u root db_container sh -c \
```
ben-palmer-sociomantic: Can we not read these values from somewhere else in the config rather than hard coding it?

andreas-bok-sociomantic: Yeah, true, they are defined as env variables in the travis build environment already. So we can read them from there

andreas-bok-sociomantic:
Updated based on comments from ben

ben-palmer-sociomantic:
LGTM

PR #235, opened 2017-09-15
Make path to generated deb packages version independent 

andreas-bok-sociomantic:
Fixes an issue where the location of generated deb 
packages depended on the version of the library


andreas-bok-sociomantic:
This turned out not to be a problem anyway. Closing

PR #236, opened 2017-09-15
Enable unittests to not return EXIT_FAILURE on SKIP

andreas-bok-sociomantic:
Add env var DRIZZLE_TEST_EXIT_ERROR_ON_SKIP
to set whether a unittest should return EXIT_FAILURE
when it is skipped


andreas-bok-sociomantic:
Closing since the changes are made obsolete by changes in #238

PR #237, opened 2017-09-15
Connect unittest: Use ENV variables for valid conn args

andreas-bok-sociomantic:
Using hardcoded value 'localhost' as host does
not work with docker-compose since the database
container is linked with hostname 'database'


andreas-bok-sociomantic:
Closing as it is part of #238

PR #238, opened 2017-09-18
Use docker-compose on travis for linux builds

andreas-bok-sociomantic:
The PR changes the existing Travis CI configuration and 
scripts to use docker-compose for all linux based builds.

The main reason is the ability to build and deploy packages 
for Ubuntu Xenial, which is not possible in the native 
travis build environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,26 +3,20 @@ language: cpp
 ## Enable cache for dependencies and directories
 cache: ccache
 
-## Enable sudo
+## Enable sudo (set explicitly although environment's with docker is sudo
+#  enabled by default
 sudo: required
 
 ## Operating Systems to use for builds
 os:
   - linux
 
-## Compilers to use for builds
-compiler:
-  - gcc
-  - clang
-
 ## Ensures the virtualization environment runs as Ubuntu 14.04 Trusty
 dist: trusty
-
 group: edge
 
 ## Services started on the build environment
 services:
- - mysql
  - docker
 
 ## Environment variables
@@ -31,58 +25,55 @@ services:
 #  build matrix, i.e. every build configuration.
 #  In turn, variables defined in the 'matrix' section each triggers a separate
 #  build.
-#
-#  Please note that test fails on first error as default
-#  To skip errors and fail silently set environment variable:
-#    SKIP_TEST_ON_ERROR=1
-#
-#  To speed up installation of brew formulas on osx auto update for homebrew is
-#  turned off by setting HOMEBREW_NO_AUTO_UPDATE=1
 env:
   global:
-    - MYSQL_SERVER=127.0.0.1
+    - COMPILER_VERSION=
+    - DEPLOY_PACKAGE=false
+    - DIST_NAME=ubuntu
+    - DIST_VERSION=xenial
+    - DOCKER_COMPOSE_VERSION=1.16.1
+    - DRIZZLE_TEST_EXIT_ERROR_ON_SKIP=0
+    - MAKE_TARGET=check
     - MYSQL_PASSWORD=''
-    - MYSQL_USER=root
     - MYSQL_PORT=3306
-    - DIST_PACKAGE_TARGET=DEB
-    - SKIP_TEST_ON_ERROR=0
+    - MYSQL_SERVER=127.0.0.1
+    - MYSQL_USER=root
     - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
+    - SKIP_TEST_ON_ERROR=0
   matrix:
-    - DIST_PACKAGE_TARGET=DEB
-    - DIST_PACKAGE_TARGET=RPM
+    - COMPILER_VERSION=4.9 MAKE_TARGET="$MAKE_TARGET:deb" DEPLOY_PACKAGE=true
+    - CXX=clang++ CC=clang COMPILER_VERSION=3.9
+    - DIST_NAME=centos DIST_VERSION=7 MAKE_TARGET=rpm
 
 ## Build matrix
 #
-#  The 'os', 'compiler' and 'env' settings are expanded into 3 jobs:
+#  Travis expands the environment variables into 6¹ jobs:
+#  All linux based builds are run in docker containers using docker-compose
 #
-#  | os     | compiler    | DIST_PACKAGE_TARGET |
+#  | os     | compiler    | make targets        |
 #  |--------|-------------|---------------------|
-#  | linux  | clang       | DEB                 |
-#  | linux  | gcc         | DEB                 |
-#  | linux* | gcc         | RPM                 |
+#  | linux  | clang       | check               |
+#  | linux  | gcc         | check deb           |
+#  | linux  | gcc         | rpm                 |
 #  | osx    | clang 6.1.0 |                     |
 #  | osx    | clang 7.3.0 |                     |
 #  | osx    | clang 8.1.0 |                     |
 #
 #  For osx, the compiler reported is the Apple LLVM version
 #
-#  * the RPM package is built in a docker container running centos7
+#  ¹ Due to excessive build times for osx only XCode 7.3.1 has been enabled.
 matrix:
   fast_finish: true
-  exclude:
-    - os: linux
-      compiler: clang
-      env: DIST_PACKAGE_TARGET=RPM
   include:
-    - os: osx
-      osx_image: xcode6.4
-      compiler: clang
+    # - os: osx
+    #   osx_image: xcode6.4
+    #   compiler: clang
     - os: osx
       osx_image: xcode7.3.1
       compiler: clang
-    - os: osx
-      osx_image: xcode8.3
-      compiler: clang
+    # - os: osx
+    #   osx_image: xcode8.3
+    #   compiler: clang
 
 ## Install dependencies
 before_install:
@@ -96,12 +87,14 @@ before_script:
 script:
   - ./travis_configure.sh run_tests
 
-# Deploy debian packages to bintray when building for a git tag
+## Deploy debian packages to bintray
+##
+## deb package built on ubuntu xenial with gcc
 deploy:
     provider: script
     script: beaver bintray upload -N pkg/deb/*.deb
     skip_cleanup: true
     on:
-        tags: true
-        repo: sociomantic-tsunami/libdrizzle-redux
-        condition: $TRAVIS_OS_NAME = 'linux' && $CC = 'gcc' && $DIST_PACKAGE_TARGET = 'DEB'
+        tags: true # must be a git tag
+        repo: sociomantic-tsunami/libdrizzle-redux # must be upstream
+        condition: $TRAVIS_OS_NAME = 'linux' && $DIST_NAME = 'ubuntu' && $DIST_VERSION = 'xenial' && $CC = 'gcc' && $MAKE_TARGET =~ 'deb' && $DEPLOY_PACKAGE = 'true'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,50 @@
+# Configuration file to run ci tests in docker containers
+#
+# Defines a database container with MySQL and a container with the environment
+# to build libdrizzle-redux. The latter container runs either ubuntu or centos
+# depending on the travis build configuration.
+#
+# See https://docs.docker.com/compose/overview/ for more information about
+# the format of the docker compose config file
+version: '3'
+
+services:
+  db:
+    image: mysql:5.7
+    restart: always
+    env_file:
+      - ./docker/mysql.env
+    container_name: mysql-container
+    volumes:
+      - "mysql_socket:/tmp/"
+      - ./docker/mysqld_binlog.cnf:/etc/mysql/mysql.conf.d/mysqld_binlog.cnf
+
+  libdrizzle:
+    env_file:
+      - ./docker/mysql.env
+      - ./docker/libdrizzle-redux.env
+    environment:
+    # environment variables interpolated from travis
+      COMPILER_VERSION: $COMPILER_VERSION
+      MAKE_TARGET: $MAKE_TARGET
+      DIST_NAME: $DIST_NAME
+      CXX: $CXX
+      CC: $CC
+    build:
+      context: ./
+    depends_on:
+      - db
+    container_name: libdrizzle-redux-container
+    links:
+      - db:database
+    volumes:
+      - "mysql_socket:/tmp/"
+      - ./:/home/libdrizzle-redux/
+      - ./docker/container_entrypoint.sh:/usr/local/bin/container_entrypoint.sh
+    working_dir: /home/libdrizzle-redux/
+    entrypoint:
+      - container_entrypoint.sh
+
+# volume used to share mysql socket between containers
+volumes:
+  mysql_socket: {}
--- a/docker/Dockerfile.centos.7
+++ b/docker/Dockerfile.centos.7
@@ -1,34 +1,8 @@
-# Docker build file to create an image running centos7 with MySQL and build
-# tools installed
-
-FROM centos/mysql-56-centos7:latest
+# Docker build file to create an image with centos 7
+FROM centos:7
 MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
 
-# Set environment variables
-ENV MYSQL_USER=user \
-    MYSQL_PASSWORD=pass \
-    MYSQL_ROOT_PASSWORD=pass \
-    MYSQL_SERVER=127.0.0.1 \
-    MYSQL_PORT=3306 \
-    MYSQL_DATABASE=test \
-    MYSQL_MASTER_USER=user \
-    MYSQL_MASTER_PASSWORD=pass \
-    MYSQL_BINLOG_FORMAT=ROW \
-    SKIP_TEST_ON_ERROR=0
-
-# Change user to `root` and install required packages to build libdrizzle-redux.
-USER root
-RUN yum install -y deltarpm
-RUN yum install -y rpm-build zlib-devel libtool openssl \
+# install required packages
+RUN yum install -y deltarpm && \
+yum install -y rpm-build zlib-devel libtool openssl \
     openssl-devel automake autoconf gcc-c++
-
-# Add host directory to container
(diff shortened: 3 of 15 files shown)


tests/unit/connect_uds.c
```diff
@@ -49,6 +49,8 @@ int main(int argc, char *argv[])
   (void)argc;
   (void)argv;
 
+  SKIP_IF(getenv("MYSQL_DOCKER_CONTAINER") != NULL);
```
leandro-lucarella-sociomantic: It fails how? Can it be fixed? I wouldn't disable tests unless there is no other way.

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+
+services:
+  db:
+    image: mysql:5.7
+    restart: always
+    environment:
+      MYSQL_ROOT_PASSWORD: pass
+      MYSQL_DATABASE: test
+      MYSQL_USER: user
+      MYSQL_PASSWORD: pass
```
leandro-lucarella-sociomantic: Why this doesn't use the `.env` file? DRY.

docker/.env
```diff
@@ -0,0 +1,11 @@
+# Docker container environment variables
+# The file is used for the docker-compose.yml::env_file setting
```
leandro-lucarella-sociomantic: BTW, why `.env` and not just `env` or `docker-compose.env`? Why hidden?

docker/docker-compose.ubuntu.xenial.yml
```diff
@@ -0,0 +1,7 @@
+# Docker compose file to override configuration for centos based images
```
leandro-lucarella-sociomantic, line 1: Is this a standard override mechanism or how is this file used?

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+# Configuration file to run ci tests in docker containers
```
leandro-lucarella-sociomantic, line 1: I would include the changes needed to use these files as part of the same commit. It is really quite complicated to figure out the interaction between these files or how will they be used in this commit.

docker/Dockerfile.centos.7
```diff
@@ -1,34 +1,12 @@
-# Docker build file to create an image running centos7 with MySQL and build
-# tools installed
-
-FROM centos/mysql-56-centos7:latest
+# Docker build file to create an image with centos 7
+FROM centos:7
```
leandro-lucarella-sociomantic, line 2: Same as previous commit, I can't see how these new files are used. I guess they aren't so far, which is weird.

docker/docker-compose.centos.7.yml
```diff
@@ -0,0 +1,7 @@
+# Docker compose file to override configuration for centos based images
```
leandro-lucarella-sociomantic, line 1: Same question about this override file.

.travis.yml
```diff
@@ -4,20 +4,14 @@ language: cpp
 cache: ccache
 
 ## Enable sudo
-sudo: required
+sudo: false
```
leandro-lucarella-sociomantic: Whot? Why isn't sudo needed when using docker compose? According to https://docs.travis-ci.com/user/docker/ you need `sudo: required` to run docker.

.travis.yml
```diff
@@ -96,12 +87,18 @@ before_script:
...
+after_script:
+  - ./travis_configure.sh after_script
+
+## Deploy debian packages to bintray
+##
+## deb package built on ubuntu xenial with gcc
 deploy:
     provider: script
-    script: beaver bintray upload -N pkg/deb/*.deb
+    script: beaver bintray upload -o -N pkg/deb/*.deb
```
leandro-lucarella-sociomantic: Is this really related to this commit?

.travis.yml
```diff
@@ -96,12 +87,18 @@ before_script:
...
     skip_cleanup: true
     on:
-        tags: true
-        repo: sociomantic-tsunami/libdrizzle-redux
-        condition: $TRAVIS_OS_NAME = 'linux' && $CC = 'gcc' && $DIST_PACKAGE_TARGET = 'DEB'
+        tags: true # must be a git tag
+        repo: sociomantic-tsunami/libdrizzle-redux # must be upstream
+        condition: $TRAVIS_OS_NAME.$DIST_NAME.$DIST_VERSION = 'linux.ubuntu.xenial' && $CC = 'gcc' && $MAKE_TARGET =~ 'deb' && $DEPLOY_PACKAGE == 'true'
```
leandro-lucarella-sociomantic: Why concatenating `$TRAVIS_OS_NAME.$DIST_NAME.$DIST_VERSION` but not `$CC`, `$MAKE_TARGET` and `$DEPLOY_PACKAGE`?

travis_configure.sh
```diff
@@ -15,27 +15,30 @@ print_error_msg ()
...
 #   - sed, (libtool), openssl, mysql
 # Returns 0
 before_install()
 {
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-        if [[ "$DIST_PACKAGE_TARGET" == "DEB" ]]; then
-            gem install fpm -v 1.8.1
+        # build images
+        docker-compose -f ./docker-compose.yml \
+            -f ./docker/docker-compose.$DIST_NAME.$DIST_VERSION.yml \
```
leandro-lucarella-sociomantic, line 37: OK, I see now, nice trick. I think the 3 last commits (including this one) should be squashed into one.

travis_configure.sh
```diff
@@ -15,27 +15,30 @@ print_error_msg ()
...
+            build
 
+        # jfrog dependency
+        if [[ "$MAKE_TARGET" =~ "deb" ]]; then
             if [[ -n "$TRAVIS_TAG" && $TRAVIS_REPO_SLUG == "sociomantic-tsunami/libdrizzle-redux" ]]; then
                 curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
```
leandro-lucarella-sociomantic, line 43: Maybe you want to use `sociomantictsunami/base` image now (it comes with `jfrog`)? Or was there a problem with that?

travis_configure.sh
```diff
@@ -148,13 +131,26 @@ run_tests()
...
 }
 
+# Invoked by the travis hook script
+#
+# - if the is run in docker stop containers and remove containers, networks,
+#   volumes, and images created by `up`.
+after_script()
+{
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        docker-compose down
```
leandro-lucarella-sociomantic: Even when this might be nice, I guess it's not really necessary (at least for travis), the VM will be killed anyway.

leandro-lucarella-sociomantic (commented):
Generally speaking, ME LIKES where this is going!

tests/unit/connect_uds.c
andreas-bok-sociomantic: It is a workaround, since I couldn't get the connection to the db container working using a socket. 
But true it should of course be fixed before it is merged.

.travis.yml
andreas-bok-sociomantic: I see. Strangely it builds with `sudo: false`. With the example they use `sudo` is required but AFAICS only because of the `sudo apt` commands. The only sudo command which runs in the native environment is the [`sudo cp /tmp/jfrog /usr/local/bin/jfrog ;` ](https://github.com/andreas-bok-sociomantic/libdrizzle-redux/blob/acdd40467a897e1a3d0eb144a9ac35d783ca7d81/travis_configure.sh#L45)
The sudo command in [`enable_mysqlbinlog()`]() is run in the OSX virtual environment but by the CI service MacStadium, https://www.macstadium.com/blog/how-travis-ci-halved-boot-times-for-os-x-and-ios-testing-at-macstadium/
Furthermore having `sudo: false` makes the build start in a container. https://docs.travis-ci.com/user/migrating-from-legacy/
However the bottleneck is still the OSX build so overall it doesn't make the whole build finish faster

.travis.yml
andreas-bok-sociomantic: Nope. Will remove, but quite convenient for testing since otherwise you would to change the git `tag` each time you wanna test the deployment

.travis.yml
andreas-bok-sociomantic: You mean making one long string to compare with? e.g. 

    $TRAVIS_OS_NAME.$DIST_NAME.$DIST_VERSION.$CC.$MAKE_TARGET.$DEPLOY_PACKAGE =~ 'linux.ubuntu.xenial.gcc.deb.true'

travis_configure.sh
andreas-bok-sociomantic: I guess I could, but all the **D** libs and packages which are installed would be obsolete for this case. 
Also currently the deployment step is handled by `travis` as defined in the `.travis.yml` file. So not much seems to be gained by using `sociomantictsunami/base`. 
However we could consider adding these Dockerfiles to the `sociomantictsunami` repo at `hub.docker.com`

travis_configure.sh
andreas-bok-sociomantic: True, I will try to test if it all the same without `docker-compose down`

docker-compose.yml
andreas-bok-sociomantic: It should. Some env variables are not relevant for the db container. Multiple files can be specified for `env_file` so will separate the vars into shared and container specific.

docker/.env
andreas-bok-sociomantic: Haha, just a consequence of following the example at docker.com blindly.

docker/docker-compose.ubuntu.xenial.yml
andreas-bok-sociomantic: Is this a standard override mechanism or how is this file used?

Yes, cf. https://docs.docker.com/compose/extends/#example-use-case

But guess the files could be generated dynamically given that they only specify which Dockerfile to use

travis_configure.sh
andreas-bok-sociomantic: Sure, guess I was more focused on the extent than the logics of the changes.

.travis.yml
leandro-lucarella-sociomantic: Have you seen the link I posted? :)

I think this is it:

> Travis CI automatically routes builds to run on Trusty sudo: required when services: docker is configured. We do not currently support use of Docker on OS X.

I wonder why just before that note they say you should use `sudo: required` if in the next line they say is automatically done when `services: - docker` is used :thinking:

leandro-lucarella-sociomantic: Yup, it looks like it's a sudo: https://travis-ci.org/sociomantic-tsunami/libdrizzle-redux/jobs/276921313#L125

.travis.yml
leandro-lucarella-sociomantic: Yes (with `=`, not `=~`). Actually I think is safer to compare every string individually, as if you have weird combinations the concatenation could match when the individual parts wont. I think is unlikely enough to be acceptable, but I really don't see any reason not to compare every variable individually instead.

travis_configure.sh
leandro-lucarella-sociomantic: The `base` image doesn't have any D stuff :)

andreas-bok-sociomantic: Ah, got things confused a bit after looking at cachalot 😀

andreas-bok-sociomantic:
Updated based on comments from @leandro-lucarella-sociomantic

.travis.yml
```diff
@@ -3,21 +3,16 @@ language: cpp
 ## Enable cache for dependencies and directories
 cache: ccache
 
-## Enable sudo
-sudo: required
+## Enable sudo (set explicitly although environment's with docker is sudo
+#  enabled  by default
+sudo: true
```
leandro-lucarella-sociomantic: Why not `required` as it was? :)

docker/Dockerfile.ubuntu.xenial
```diff
@@ -0,0 +1,35 @@
+# Dockerfile for building an image with ubuntu xenial
+FROM sociomantictsunami/base
+MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
+
+WORKDIR /root
```
leandro-lucarella-sociomantic: :scream:

docker/Dockerfile.ubuntu.xenial
```diff
@@ -0,0 +1,35 @@
...
+
+# install apt-fast to download required packages in parallel
+RUN echo "deb http://ppa.launchpad.net/saiarcot895/myppa/ubuntu xenial main" >> /etc/apt/sources.list
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DC058F40
+RUN apt-get update
+RUN apt-get install --no-install-recommends -y apt-fast
+
+# install required packages
+RUN apt-fast -y install --no-install-recommends \
+    libssl-dev autoconf libtool make automake zlib1g-dev
```
leandro-lucarella-sociomantic: Every `RUN` command in Docker creates a new layer, it is usually recommended to create some script and just use one `RUN` (or if it's just a few commands use `RUN cmd && cmd && cmd`). Not a big deal, just FYI.

leandro-lucarella-sociomantic (commented):
Just a few more comments, but changes look good in general

.travis.yml
andreas-bok-sociomantic: Ah sure, should be `required`

docker/Dockerfile.ubuntu.xenial
andreas-bok-sociomantic: hehe, will change to a less contested dir

docker/Dockerfile.ubuntu.xenial
andreas-bok-sociomantic: Oki, you mean a new temp image will be created for each RUN cmd? 

I will wrap the cmds  in a script

docker/Dockerfile.ubuntu.xenial
leandro-lucarella-sociomantic: Afterwards you change it to `libdrizzle-redux` or something. I wonder why you set it at all.

docker/Dockerfile.ubuntu.xenial
leandro-lucarella-sociomantic: Yup

andreas-bok-sociomantic:
Updated based on comments from @leandro-lucarella-sociomantic 
Additional changes:
- moved the contents of `install_dependencies.sh` to a function in `container_entrypoint.sh`
- removed obsolete and out-commented install of **fpm** since it is already available in `sociomantictsunami/base`

docker/Dockerfile.ubuntu.xenial
```diff
@@ -0,0 +1,10 @@
+# Dockerfile for building an image with ubuntu xenial
+FROM sociomantictsunami/base
+MAINTAINER andreas-bok-sociomantic <andreas.bok@sociomantic.com>
+
+WORKDIR /home
```
leandro-lucarella-sociomantic: I wonder why you need to set a workdir at all.

andreas-bok-sociomantic: Hmm, isn't `root` the default user in the images, which means that `$HOME` is  `/root`?

andreas-bok-sociomantic: Cf, https://docs.docker.com/engine/reference/run/#user and https://docs.docker.com/engine/reference/run/#workdir

andreas-bok-sociomantic:
Updated
- fixed issue with socket connection by adding a volume in which the mysql socket is shared between the containers

docker/Dockerfile.ubuntu.xenial
leandro-lucarella-sociomantic: When building the image it is root. But the question is why you need to know which is the work directory? Are you writing any files to it?

I think you are mixing up stuff, this is image creating, for running you just pass `-u`. The `USER` and `WORKDIR` commands make (more?) sense when using docker to actually deploy stuff, but for testing you should just run `docker run -u user -w workdir` if you want to set it to anything in particular, you don't have to do anything at the image creation unless you need it to actually create the image.

andreas-bok-sociomantic: > Are you writing any files to it?

Only writing the init script `ubuntu_xenial_image_run.sh` to `/usr/local/bin/`

> if you want to set it to anything in particular, you don't have to do anything at the image creation unless you need it to actually create the image.

Ah, I can set the working directory in the `docker-compose.yml` file, https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir

andreas-bok-sociomantic:
Updated
- removed `WORK_DIR` steps from Dockerfiles
- specify the working directory in `docker-compose.yml` instead

leandro-lucarella-sociomantic:
Is there any ETA to finalize this PR?

leandro-lucarella-sociomantic:
Also, can you add a "Fixes #239" in the commit that removes building for trusty, so it automatically gets closed when this is merged?

.travis.yml
```diff
@@ -31,58 +25,55 @@ services:
...
-#
-#  To speed up installation of brew formulas on osx auto update for homebrew is
-#  turned off by setting HOMEBREW_NO_AUTO_UPDATE=1
 env:
   global:
-    - MYSQL_SERVER=127.0.0.1
+    - COMPILER_VERSION=
+    - DEPLOY_PACKAGE=false
+    - DIST_NAME=ubuntu
+    - DIST_VERSION=xenial
```
leandro-lucarella-sociomantic, line 33: I'd mention that this commit also stops building for trusty and starts building for xenial.

andreas-bok-sociomantic:
> Is there any ETA to finalize this PR?

I will address the comments, and then if everything is ok, it will be part of the [`v5.7.1`](https://github.com/sociomantic-tsunami/libdrizzle-redux/milestone/34) release

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic addressed your comments. If all is good then let's merge! :)

leandro-lucarella-sociomantic (approved):

PR #243, opened 2017-10-13
Fix brew runtime error on travis

andreas-bok-sociomantic:
Homebrew requires Ruby 2.3 to run. The Travis images for OSX are outdated, and the only way to get Homebrew working is to run brew update before doing anything else, as suggested in https://github.com/Homebrew/brew/issues/3299


andreas-bok-sociomantic:
Thanks

PR #266, opened 2018-01-05
[TRAVIS] Use build stages in travis configuration

andreas-bok-sociomantic:
Migrate the `.travis.yml` config file to use build stages.

Besides adapting the logic the PR changes the configuration wrt.: 

- Remove **ENV** variable **DEPLOY_PACKAGE** since deployments are defined explicitly in the `deploy` step. 
- Remove the  `repo: sociomantic-tsunami/libdrizzle-redux` condition in the deployment step. Allows for testing with a developer's own **bintray** account, thus avoiding pushing to the upstream **bintray** repo. 
- only enable `docker` as service for linux builds using the `jobs.templates` config as it is not required for OSX builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,26 @@
 language: cpp
 
 ## Enable cache for dependencies and directories
-cache: ccache
+cache:
+  directories:
+    - build/travis-cache
 
-## Enable sudo (set explicitly although environment's with docker is sudo
-#  enabled by default
+## Enable sudo (set explicitly although it is enabled by default when docker is
+#  running as a service
 sudo: required
 
 ## Operating Systems to use for builds
-os:
-  - linux
+os: linux
 
 ## Ensures the virtualization environment runs as Ubuntu 14.04 Trusty
 dist: trusty
 
-## Services started on the build environment
-services:
- - docker
-
 ## Environment variables
 #
-#  The 'global' section defines variables which are added to every row in the
-#  build matrix, i.e. every build configuration.
-#  In turn, variables defined in the 'matrix' section each triggers a separate
-#  build.
+#  The 'global' section defines variables common for all jobs
 env:
   global:
     - COMPILER_VERSION=
-    - DEPLOY_PACKAGE=false
     - DIST_NAME=ubuntu
     - DIST_VERSION=xenial
     - DOCKER_COMPOSE_VERSION=1.16.1
@@ -39,60 +32,83 @@ env:
     - MYSQL_USER=root
     - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     - SKIP_TEST_ON_ERROR=0
-  matrix:
-    - COMPILER_VERSION=4.9 MAKE_TARGET="$MAKE_TARGET:deb" DEPLOY_PACKAGE=true
-    - CXX=clang++ CC=clang COMPILER_VERSION=3.9
-    - DIST_NAME=centos DIST_VERSION=7 MAKE_TARGET=rpm
+    - CACHE_DIR=build/travis-cache
 
-## Build matrix
-#
-#  Travis expands the environment variables into 6¹ jobs:
-#  All linux based builds are run in docker containers using docker-compose
-#
-#  | os     | compiler    | make targets        |
-#  |--------|-------------|---------------------|
-#  | linux  | clang       | check               |
-#  | linux  | gcc         | check deb           |
-#  | linux  | gcc         | rpm                 |
-#  | osx    | clang 6.1.0 |                     |
-#  | osx    | clang 7.3.0 |                     |
-#  | osx    | clang 8.1.0 |                     |
-#
-#  For osx, the compiler reported is the Apple LLVM version
-#
-#  ¹ Due to excessive build times for osx only XCode 7.3.1 has been enabled.
-matrix:
-  fast_finish: true
-  include:
-    # - os: osx
-    #   osx_image: xcode6.4
-    #   compiler: clang
-    - os: osx
-      compiler: clang
-    # - os: osx
-    #   osx_image: xcode8.3
-    #   compiler: clang
-
-## Install dependencies
+## Build steps for all jobs
 before_install:
   - ./travis_configure.sh before_install
-
-## Prepare build system for tests
 before_script:
   - ./travis_configure.sh before_script
-
-## Run tests, and packaging scripts
 script:
   - ./travis_configure.sh run_tests
 
-## Deploy debian packages to bintray
-##
-## deb package built on ubuntu xenial with gcc
-deploy:
-    provider: script
-    script: beaver bintray upload -D $DIST_VERSION -N build/pkg/deb/*.deb
-    skip_cleanup: true
-    on:
-        tags: true # must be a git tag
-        repo: sociomantic-tsunami/libdrizzle-redux # must be upstream
-        condition: $TRAVIS_OS_NAME = 'linux' && $DIST_NAME = 'ubuntu' && $DIST_VERSION = 'xenial' && $CC = 'gcc' && $MAKE_TARGET =~ 'deb' && $DEPLOY_PACKAGE = 'true'
+## Defines the order of the build stages.
+#
+#  The deployment stage is only done is building semver git tag
+stages:
+  - test
+  - deploy
+
+## Defines the jobs to run in each stage
+#
+#  Travis combines the variables defined in env.global with the variables
+#  defined for each job
+#
+#  The resulting build matrix is as follows:
+#
+#  | os     | dist   | compiler  | make targets | deploy |
+#  |--------|--------|-----------|--------------|--------|
+#  | linux  | ubuntu | clang     | check        | no     |
+#  | linux  | ubuntu | gcc       | check deb    | yes    |
+#  | linux  | centos | gcc       | check rpm    | yes    |
+#  | osx    | sierra | clang¹    | check        | no     |
+#
+#  Linux based builds are run in docker containers using docker-compose.
+#
+#  OSX based builds are done by a third-party provider
+#
+#  ¹ The compiler reported is the Apple LLVM version
+jobs:
+  templates:
+    # Global config for builds on linux
+    - &linux-build
+      services:
+        - docker
+    # Global config for builds on osx
+    - &osx-build
+      os: osx
+      compiler: clang
+      if: tag =~ osx$ or branch =~ osx$
+    # Global config for deployment stage
+    - &deploy
+      os: linux
+
+  include:
+    # Sets up the build environment, compiles the code and runs tests
+    - <<: *linux-build
+      env:
+        COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
+      before_cache:
+      - if [[ -n $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
+    - <<: *linux-build
+      env:
+        CXX=clang++ CC=clang COMPILER_VERSION=3.9
+    - <<: *linux-build
+      env:
+        DIST_NAME=centos DIST_VERSION=7 MAKE_TARGET=rpm
+    - <<: *osx-build
+      env:
+        DIST_NAME=osx
+
+    ## Deployment stage
+    - <<: *deploy
+      # Deploy deb packages to bintray. Doesn't build the packages but uses the
+      # cache from the first stage.
+      stage: deploy
+      if: tag IS present AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
+      env:
+        COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
+      before_install:
+        - ./travis_configure.sh install_dependency jfrog
+        - beaver bintray upload -D $DIST_VERSION -N $CACHE_DIR/*.deb
+        - exit 0
--- a/travis_configure.sh
+++ b/travis_configure.sh
@@ -22,6 +22,30 @@ print_error_msg ()
     return 0
 }
 
+install_dependency()
+{
+    if [[ -z $1 ]]; then
+        echo "Missing argument"
+        return 1
+    fi
+
+    case $1 in
+        jfrog)
+            echo "Installing $1"
+            curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
+            chmod a+x /tmp/jfrog ;
+            mkdir -p $HOME/bin ;
+            cp /tmp/jfrog $HOME/bin/jfrog ;
+            #export PATH=$HOME/bin:$PATH
+            ;;
+        *)
+            echo "installation of $1 is not supported"
+            ;;
+    esac
+
+    return 0
+}
+
 # Script which is run before the installation script is called
 #
 # For linux based builds docker-compose is used to set up the build environment
@@ -39,9 +63,7 @@ before_install()
 
         # jfrog dependency
         if [[ -n "$TRAVIS_TAG" ]]; then
-            curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
-            chmod a+x /tmp/jfrog ;
-            sudo cp /tmp/jfrog /usr/local/bin/jfrog ;
+            install_dependency jfrog
         fi
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
         brew update
@@ -136,6 +158,9 @@ elif [[ $1 == "before_script" ]]; then
     before_script
 elif [[ $1 == "run_tests" ]]; then
     run_tests
+elif [[ $1 == "install_dependency" ]];  then
+    shift
+    install_dependency $1
 else
     echo "invalid function call"
     exit 1


andreas-bok-sociomantic:
blocked: deploy step not started

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic, just wondering how you found out about the `jobs.templates` for build stages. The `templates` option doesn't seem to documented anywhere.

.travis.yml
```diff
@@ -39,60 +25,82 @@ env:
...
-      compiler: clang
-    # - os: osx
-    #   osx_image: xcode8.3
-    #   compiler: clang
 
-## Install dependencies
-before_install:
-  - ./travis_configure.sh before_install
+stages:
+  - deploy
```
leandro-lucarella-sociomantic: I guess it should be:

```
stages:
  - test
  - deploy
```

leandro-lucarella-sociomantic (commented):
The deploy stage is shown before the test stage, which is weird. I guess you need to specify the stage order?

leandro-lucarella-sociomantic:
Other than that, LGTM.

.travis.yml
andreas-bok-sociomantic: Right,

andreas-bok-sociomantic:
The fact that there is no [data persistence between stages](https://docs.travis-ci.com/user/build-stages/#Data-persistence-between-stages-and-jobs) results in an obsolete build in the deploy stage. 

Doing the deployment in the `test` stage would eliminate that but sort of beats the purpose of having the whole project build successfully before deploying. 

Though AFAICS this is what you do in the build stages for the other tsunami projects using travis

andreas-bok-sociomantic:
Another thing @leandro-lucarella-sociomantic did you look into caching the docker image so it isn't downloaded for every build? 
- https://giorgos.sealabs.net/docker-cache-on-travis-and-docker-112.html
- http://atodorov.org/blog/2017/08/07/faster-travis-ci-tests-with-docker-cache/

andreas-bok-sociomantic:
@leandro-lucarella-sociomantic it is ready for a final review. Just the osx build hanging as usual

andreas-bok-sociomantic:
> Another thing leandro-lucarella-sociomantic did you look into caching the docker image so it isn't downloaded for every build?

Cf. #272

leandro-lucarella-sociomantic:
> Doing the deployment in the test stage would eliminate that but sort of beats the purpose of having the whole project build successfully before deploying.

Other projects are just rebuilding on the deployment stage. Yes, it sucks that travis doesn't support passing artefacts.

> Another thing @leandro-lucarella-sociomantic did you look into caching the docker image so it isn't downloaded for every build?

Yes, I tried that and it was always about the same (or slower) than just rebuilding the image: https://github.com/sociomantic-tsunami/ocean/pull/103

.travis.yml
```diff
@@ -39,60 +31,87 @@ env:
...
+## Defines the order of the build stages.
+#
+#  The deployment stage is only done is building semver git tag
+stages:
+  - test
+  - name: deploy
+    if:
+      $TRAVIS_OS_NAME = 'linux' AND
+      env(DIST_NAME) = 'ubuntu' AND
+      env(DIST_VERSION) = 'xenial'
```
leandro-lucarella-sociomantic: This can actually be put in the deploy job. I think is more clear to have all in one place.

.travis.yml
```diff
@@ -39,60 +31,87 @@ env:
...
+        DIST_NAME=osx
+
+    ## Deployment stage
+    - <<: *deploy
+      stage: deploy
+      env:
+        MAKE_TARGET=deb
+      if:
+        tag IS present AND
+        tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
```
leandro-lucarella-sociomantic: And you already have a `if` here, making it even more complicated to follow. Even more, I think this `if` overrides the one you wrote before!

andreas-bok-sociomantic:
> Yes, I tried that and it was always about the same (or slower) than just rebuilding the image: sociomantic-tsunami/ocean#103

Did an implementation where dropbox is used to upload and download cache. Of course AWS would be the more natural choice, i.e. if it didn't cost money,  
 
- https://github.com/andreas-bok-sociomantic/libdrizzle-redux/commits/save-deb-packages
- https://travis-ci.org/andreas-bok-sociomantic/libdrizzle-redux/builds/327653552

leandro-lucarella-sociomantic:
Right, no, we didn't try only to store the `.deb` packages, but to cache the docker images, I thought you were talking about that judging from the links you sent (which seems to be just pushing to the docker registry, which is what I tried). Of course to only pass artefacts it would be more efficient, but yeah, we need a files provider for that.

I wanted to see how you implemented it but your commits are.... not descriptive. Making it very hard to try to see what you did :)

leandro-lucarella-sociomantic:
I think we might be able to use the [Travis cache](https://docs.travis-ci.com/user/caching/) for this, I'm not sure how it works between stages, but it might work...

andreas-bok-sociomantic:
> I wanted to see how you implemented it but your commits are.... not descriptive. Making it very hard to try to see what you did :)

Haha, still at the development stage....

> I think we might be able to use the Travis cache for this, I'm not sure how it works between stages, but it might work...

Will take a look

leandro-lucarella-sociomantic:
Ah, you need to install jfrog and can't use beaver for deployment like this... Still I guess is something that beaver could require the user to set up the cache if it works, but we need to use travis VM to run the deployment stuff, otherwise we have to create the docker image which will take most of the time.

There is some built-in deployment support from travis, but I didn't use it because it was too verbose. But considering this new stage option, maybe we can have a look again.

leandro-lucarella-sociomantic:
The cache works to pass artefacts, so it might be an interesting option:

* https://www.travis-ci.org/leandro-lucarella-sociomantic/tangort/jobs/327819949#L499
* https://github.com/leandro-lucarella-sociomantic/tangort/commit/a8a90f77b05401a46e25094b66f011ce69eb3655

OTOH, deployment happens so infrequently that it hardly worth the effort TBH.

.travis.yml
andreas-bok-sociomantic: > Even more, I think this if overrides the one you wrote before!

Damn it, so much for templates

.travis.yml
andreas-bok-sociomantic: Realised it isn't required since the vars above in `env.global` have exactly the values I check for

andreas-bok-sociomantic:
> The cache works to pass artefacts, so it might be an interesting option:

Can't get caching to work between stages. So will revert to the easy option of building the deb again.

andreas-bok-sociomantic:
Addressed comments 

- Removed condition for `deploy` stage since it it executed with the `env.global` vars.
- Removed the check for **BINTRAY_USER** and **BINTRAY_KEY** in the `deploy` step since `beaver` will report an error if these env vars are missing 

Previous head was 2bfa1ac, new **HEAD** at eed5613.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 13 ++-----------
 1 file changed, 2 insertions(+), 11 deletions(-)

diff --git a/.travis.yml b/.travis.yml
index 9664d01..dd770cf 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,11 +45,7 @@ script:
 #  The deployment stage is only done is building semver git tag
 stages:
   - test
-  - name: deploy
-    if:
-      $TRAVIS_OS_NAME = 'linux' AND
-      env(DIST_NAME) = 'ubuntu' AND
-      env(DIST_VERSION) = 'xenial'
+  - deploy

 ## Defines the jobs to run in each stage
 #
@@ -102,16 +98,11 @@ jobs:
     ## Deployment stage
     - <<: *deploy
       stage: deploy
+      if: tag IS present AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
       env:
         MAKE_TARGET=deb
-      if:
-        tag IS present AND
-        tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
       ## Deploy debian packages to bintray
       deploy:
         - provider: script
           script: beaver bintray upload -D $DIST_VERSION -N build/pkg/deb/*.deb
           skip_cleanup: true
-          on:
-            all_branches: true
-            condition: $BINTRAY_USER != '' AND $BINTRAY_KEY != ''

```
</details>

andreas-bok-sociomantic:
> Can't get caching to work between stages. So will revert to the easy option of building the deb again.

Works now when the env variables are identical, as you mentioned

.travis.yml
```diff
@@ -39,60 +31,78 @@ env:
...
+      env:
+        DIST_NAME=centos DIST_VERSION=7 MAKE_TARGET=rpm
+    - <<: *osx-build
+      env:
+        DIST_NAME=osx
+
+    ## Deployment stage
+    - <<: *deploy
+      stage: deploy
+      if: tag IS present AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
```
leandro-lucarella-sociomantic: Why the regex on the tag?

andreas-bok-sociomantic: So only `semver` style tags triggers a deployment. Means you can test git tags upstream without deploying to `bintray `

andreas-bok-sociomantic:
Added caching of built artifacts from the first stage 

This allows us to skip building the packages in the 
deployment and simply deploy the cached packages 

Previous head was bdb6d2e, new **HEAD** at 2bfa1ac.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml         | 22 ++++++++++++++--------
 travis_configure.sh | 31 ++++++++++++++++++++++++++++---
 2 files changed, 42 insertions(+), 11 deletions(-)

diff --git a/.travis.yml b/.travis.yml
index dd770cf..0e4b44c 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: cpp
 
 ## Enable cache for dependencies and directories
-cache: ccache
+cache:
+  directories:
+    - $CACHE_DIR
 
 ## Enable sudo (set explicitly although it is enabled by default when docker is
 #  running as a service
@@ -31,6 +33,7 @@ env:
     - MYSQL_USER=root
     - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     - SKIP_TEST_ON_ERROR=0
+    - CACHE_DIR=build/travis-cache/
 
 ## Build steps for all jobs
 before_install:
@@ -84,7 +87,9 @@ jobs:
     # Sets up the build environment, compiles the code and runs tests
     - <<: *linux-build
       env:
-        COMPILER_VERSION=4.9
+        COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
+      before_cache:
+      - cp build/pkg/deb/*.deb build/travis-cache/
     - <<: *linux-build
       env:
         CXX=clang++ CC=clang COMPILER_VERSION=3.9
@@ -97,12 +102,13 @@ jobs:
 
     ## Deployment stage
     - <<: *deploy
+      # Deploy deb packages to bintray. Doesn't build the packages but uses the
+      # cache from the first stage.
       stage: deploy
       if: tag IS present AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+
       env:
-        MAKE_TARGET=deb
-      ## Deploy debian packages to bintray
-      deploy:
-        - provider: script
-          script: beaver bintray upload -D $DIST_VERSION -N build/pkg/deb/*.deb
-          skip_cleanup: true
+        COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
+      before_install:
+        - ./travis_configure.sh install_dependency jfrog
+        - beaver bintray upload -D $DIST_VERSION -N $CACHE_DIR/*.deb
+        - exit 0
diff --git a/travis_configure.sh b/travis_configure.sh
index 4823a50..ad4b840 100755
--- a/travis_configure.sh
+++ b/travis_configure.sh
@@ -22,6 +22,30 @@ print_error_msg ()
     return 0
 }
 
+install_dependency()
+{
+    if [[ -z $1 ]]; then
+        echo "Missing argument"
+        return 1
+    fi
+
+    case $1 in
+        jfrog)
+            echo "Installing $1"
+            curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
+            chmod a+x /tmp/jfrog ;
+            mkdir -p $HOME/bin ;
+            cp /tmp/jfrog $HOME/bin/jfrog ;
+            #export PATH=$HOME/bin:$PATH
+            ;;
+        *)
+            echo "installation of $1 is not supported"
+            ;;
+    esac
+
+    return 0
+}
+
 # Script which is run before the installation script is called
 #
 # For linux based builds docker-compose is used to set up the build environment
@@ -39,9 +63,7 @@ before_install()
 
         # jfrog dependency
         if [[ -n "$TRAVIS_TAG" ]]; then
-            curl -XGET -L -k 'https://api.bintray.com/content/jfrog/jfrog-cli-go/$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64' > /tmp/jfrog ;
-            chmod a+x /tmp/jfrog ;
-            sudo cp /tmp/jfrog /usr/local/bin/jfrog ;
+            install_dependency jfrog
         fi
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
         brew update
@@ -136,6 +158,9 @@ elif [[ $1 == "before_script" ]]; then
     before_script
 elif [[ $1 == "run_tests" ]]; then
     run_tests
+elif [[ $1 == "install_dependency" ]];  then
+    shift
+    install_dependency $1
 else
     echo "invalid function call"
     exit 1

```
</details>

leandro-lucarella-sociomantic:
You are losing the ccache cache then. I guess that's bad (it will make *all* the builds slower).

andreas-bok-sociomantic:
> You are losing the ccache cache then. I guess that's bad (it will make all the builds slower).

Well, IIUC the `ccache` won't help in the `Test` stage since the builds run in parallel. 
Of course jobs in the `Deploy` stage won't be able to utilize the `ccache`. But they only run for tagged builds (and changes to the main branches for the coversalls job #271)

leandro-lucarella-sociomantic:
But you are killing the cache cache, not the inter-stage cache. The cache was created before the stages were even implemented and it's to speed up the building of each build. Suppose that this project compiles 100 C files and you build it so the `.o` are cached by cache. Then you commit a new change that changes only one C file, so only one `.o` file has to be regenerated and the project only needs to be linked again. That saves a lot of time.

leandro-lucarella-sociomantic:
Have a look at Travis cache documentation :)

leandro-lucarella-sociomantic:
I don't know if it's possible, but you should cache both (and the package probably only when you are building a tag, otherwise is just a waste of time to cache it).

andreas-bok-sociomantic:
> You are losing the ccache cache then. I guess that's bad (it will make all the builds slower).

> I don't know if it's possible, but you should cache both (and the package probably only when you are building a tag, otherwise is just a waste of time to cache it).

But all builds are done directly in the docker containers. So the real speed up would be if these could be cached and shared between builds/stages. 

On the barebone `travis` build environment no dependencies are installed except for `jfrog`

andreas-bok-sociomantic:
I don't see any degradation of build time after removing `ccache`. And since `sudo` is enabled when `docker` runs as service the builds can't run in the **container-based** environment, which makes the start up time considerable slower, https://docs.travis-ci.com/user/reference/overview/

andreas-bok-sociomantic:
> You are losing the ccache cache then. I guess that's bad (it will make all the builds slower).

But won't need to copy to a custom directory, should be able to just use the `ccache` directory `$HOME/.ccache`

andreas-bok-sociomantic:
Use the default `ccache` directory $HOME/.ccache
to cache artifacts 

Previous head was 10ad6eb, new **HEAD** at bdb6d2e.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 10 ++++------
 1 file changed, 4 insertions(+), 6 deletions(-)

diff --git a/.travis.yml b/.travis.yml
index 0e4b44c..8f18049 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: cpp
 
 ## Enable cache for dependencies and directories
-cache:
-  directories:
-    - $CACHE_DIR
+cache: ccache
 
 ## Enable sudo (set explicitly although it is enabled by default when docker is
 #  running as a service
@@ -33,7 +31,7 @@ env:
     - MYSQL_USER=root
     - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     - SKIP_TEST_ON_ERROR=0
-    - CACHE_DIR=build/travis-cache/
+    - CCACHE_DIR=$HOME/.ccache
 
 ## Build steps for all jobs
 before_install:
@@ -89,7 +87,7 @@ jobs:
       env:
         COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
       before_cache:
-      - cp build/pkg/deb/*.deb build/travis-cache/
+      - cp build/pkg/deb/*.deb $CCACHE_DIR/
     - <<: *linux-build
       env:
         CXX=clang++ CC=clang COMPILER_VERSION=3.9
@@ -110,5 +108,5 @@ jobs:
         COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
       before_install:
         - ./travis_configure.sh install_dependency jfrog
-        - beaver bintray upload -D $DIST_VERSION -N $CACHE_DIR/*.deb
+        - beaver bintray upload -D $DIST_VERSION -N $CCACHE_DIR/*.deb
         - exit 0

```
</details>

andreas-bok-sociomantic:
Previous head was 87f5861, new **HEAD** at 10ad6eb.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)

diff --git a/.travis.yml b/.travis.yml
index 8f18049..9669906 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ cache: ccache
 sudo: required
 
 ## Operating Systems to use for builds
-os:
-  - linux
+os: linux
 
 ## Ensures the virtualization environment runs as Ubuntu 14.04 Trusty
 dist: trusty
@@ -87,7 +86,7 @@ jobs:
       env:
         COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
       before_cache:
-      - cp build/pkg/deb/*.deb $CCACHE_DIR/
+      - if [[ -z $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CCACHE_DIR/; fi
     - <<: *linux-build
       env:
         CXX=clang++ CC=clang COMPILER_VERSION=3.9

```
</details>

andreas-bok-sociomantic:
Add conditions for osx build

Previous head was e0258d0, new **HEAD** at 87f5861.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 1 +
 1 file changed, 1 insertion(+)

diff --git a/.travis.yml b/.travis.yml
index 9669906..a4f8030 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,7 @@ jobs:
     - &osx-build
       os: osx
       compiler: clang
+      if: tag =~ osx$ or branch =~ osx$
     # Global config for deployment stage
     - &deploy
       os: linux

```
</details>

leandro-lucarella-sociomantic:
> But all builds are done directly in the docker containers. So the real speed up would be if these could be cached and shared between builds/stages.

Right, to get advantage of `ccache` you probably need to install it in the docker image and actually use it when bulding, as well as mounting the directory where the cache is actually stored.

OK then, I guess ccache cache was configured but not really used.

leandro-lucarella-sociomantic:
I think the cache in these last commits is a bit of a mess. If you keep using the ccache cache, I would store it in the default dir, and just cache both `.ccache` and `travis-cache` directories, there is no need to mix both.

andreas-bok-sociomantic:
> I think the cache in these last commits is a bit of a mess. If you keep using the ccache cache, I would store it in the default dir, and just cache both .ccache and travis-cache directories, there is no need to mix both.

Not sure I understand. `$HOME/.ccache`is the default `ccache` [dir ](https://docs.travis-ci.com/user/caching#ccache-cache) 

**CCACHE_DIR=$HOME/.ccache** is just to have it available as an env variable.

andreas-bok-sociomantic:
> OK then, I guess ccache cache was configured but not really used.

Yes. But used now after ensuring that the build configuration for the deb build and deployment are identical.

andreas-bok-sociomantic:
Fixed wrong check for TRAVIS_TAG 

Previous head was aa93083, new **HEAD** at 5af9c66.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/.travis.yml b/.travis.yml
index 4d28efc..fb4717b 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ jobs:
       env:
         COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
       before_cache:
-      - if [[ -z $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
+      - if [[ ! -z $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
     - <<: *linux-build
       env:
         CXX=clang++ CC=clang COMPILER_VERSION=3.9

```
</details>

.travis.yml
```diff
@@ -39,60 +32,83 @@ env:
...
+    - &deploy
+      os: linux
+
+  include:
+    # Sets up the build environment, compiles the code and runs tests
+    - <<: *linux-build
+      env:
+        COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
+      before_cache:
+      - if [[ ! -z $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
```
leandro-lucarella-sociomantic: `! -z` == `-n` :)

andreas-bok-sociomantic:
use `-n` instead of _unsafe_ `! -z`

Previous head was 295fb27, new **HEAD** at aa93083.

<details>
<summary>Diff between heads are:</summary>

```diff

 .travis.yml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/.travis.yml b/.travis.yml
index fb4717b..9ec1dda 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ jobs:
       env:
         COMPILER_VERSION=4.9 MAKE_TARGET=$MAKE_TARGET:deb
       before_cache:
-      - if [[ ! -z $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
+      - if [[ -n $TRAVIS_TAG ]]; then cp build/pkg/deb/*.deb $CACHE_DIR/; fi
     - <<: *linux-build
       env:
         CXX=clang++ CC=clang COMPILER_VERSION=3.9

```
</details>

andreas-bok-sociomantic:
Want to add the changes into `v5.7.x` first. Won't be used but keeping it until until #275 has been merged.

andreas-bok-sociomantic:
closed after #275 was merged into branch `v6.1.x`

PR #268, opened 2018-01-06
[TRAVIS] Simplify conditions for jfrog script download

andreas-bok-sociomantic:
Only download 'jfrog' for tagged builds where it
may be required to upload packages to bintray

### DIFF
--- a/travis_configure.sh
+++ b/travis_configure.sh
@@ -38,12 +38,10 @@ before_install()
             build
 
         # jfrog dependency
-        if [[ "$MAKE_TARGET" =~ "deb" ]]; then
-            if [[ -n "$TRAVIS_TAG" && $TRAVIS_REPO_SLUG == "sociomantic-tsunami/libdrizzle-redux" ]]; then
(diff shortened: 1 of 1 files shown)


travis_configure.sh
```diff
@@ -38,12 +38,10 @@ before_install()
             build
 
         # jfrog dependency
-        if [[ "$MAKE_TARGET" =~ "deb" ]]; then
-            if [[ -n "$TRAVIS_TAG" && $TRAVIS_REPO_SLUG == "sociomantic-tsunami/libdrizzle-redux" ]]; then
```
ben-palmer-sociomantic, line 42: So you don't need the check on `TRAVIS_REPO_SLUG` any more?

andreas-bok-sociomantic: The check was done to only install `jfrog` for upstream builds. But while testing it is quite convenient to be able to test deployment for your own repo, cf. the description for #266

ben-palmer-sociomantic:
LGTM

PR #273, opened 2018-01-12
[TRAVIS] Modify installation of additional dependencies

andreas-bok-sociomantic:
Add a separate function for installing artifacts in the build environment

Remove the sudo dependency when installing `jfrog` by copying the binary to `$HOME/bin` and add this
to **PATH**


andreas-bok-sociomantic:
Closed in lieu of #266

PR #274, opened 2018-01-12
[TRAVIS] Minor fixes to contrainer_entrypoint script

andreas-bok-sociomantic:
Fix invalid check for clang compiler

Don't include recommended packages when installing the
compiler


andreas-bok-sociomantic:
Fails because of osx build timeout...so will merge anyhow

PR #282, opened 2018-02-05
Add coverage for API function drizzle_free_context

andreas-bok-sociomantic:


andreas-bok-sociomantic:
Rebased on latest v6.x.x

tests/unit/event_callback.c
```diff
@@ -81,6 +81,16 @@ extern drizzle_return_t drizzle_event_callback(drizzle_st *con, short events,
   return DRIZZLE_RETURN_OK;
 }
 
+extern void drizzle_free_context(drizzle_st *con, void *context);
+extern void drizzle_free_context(drizzle_st *con, void *context)
+{
+  (void)con;
```
mathias-lang-sociomantic, line 87: Why do you specify the name then ?

andreas-bok-sociomantic: Otherwise it results in an `unused variable warning`

mathias-lang-sociomantic: AFAIK compilers won't issue this warning if you don't specify a name:
```
extern void drizzle_free_context(drizzle_st*, void *context) 
```

andreas-bok-sociomantic: That results in : 

`../tests/unit/event_callback.c:85:1: error: parameter name omitted`

andreas-bok-sociomantic: compiler warning flags are on viagra

mathias-lang-sociomantic: That's why I don't want to do C++ anymore

PR #288, opened 2018-04-17
Minor fixes

federico-cuello-sociomantic:


src/ssl.cc
```diff
@@ -50,8 +50,11 @@ drizzle_return_t drizzle_set_ssl(drizzle_st *con, const char *key, const char *c
 
   if (cipher)
   {
-    drizzle_set_error(con, __FILE_LINE_FUNC__, "Cannot set the SSL cipher list");
-    return DRIZZLE_RETURN_SSL_ERROR;
+    if (SSL_CTX_set_cipher_list((SSL_CTX*)con->ssl_context, cipher) != 1)
+    {
+      drizzle_set_error(con, __FILE_LINE_FUNC__, "Cannot set the SSL cipher list");
```
andreas-bok-sociomantic, line 60: Working on a PR to include the underlying error from SSL. 
However fine to add the error message now

andreas-bok-sociomantic (commented):
With to the travis ci error.
Since it is built in a docker containers the specific openssl version to install is handled here: 

https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v6.x.x/docker/ubuntu_xenial_image_run.sh#L20

https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v6.x.x/docker/Dockerfile.centos.7#L7

src/ssl.cc
```diff
@@ -46,7 +46,7 @@
 
 drizzle_return_t drizzle_set_ssl(drizzle_st *con, const char *key, const char *cert, const char *ca, const char *capath, const char *cipher)
 {
-  con->ssl_context= SSL_CTX_new(TLSv1_client_method());
+  con->ssl_context= SSL_CTX_new(TLS_client_method());
```
andreas-bok-sociomantic: Could you add a short ci message e.g. 

> With `TLS_client_method` the actual protocol version used will be negotiated to the highest version mutually supported by the client and the server.

federico-cuello-sociomantic: Using `SSLv23_client_method()` instead and improved commit message.

andreas-bok-sociomantic:
LGTM

andreas-bok-sociomantic:
> With to the travis ci error.
Since it is built in a docker containers the specific openssl version to install is handled here:

Ah, see that the ci build fail was fixed in the code using `SSLv23_client_method` :+1:

src/pack.cc
```diff
@@ -402,7 +402,7 @@ unsigned char *drizzle_pack_auth(drizzle_st *con, unsigned char *ptr,
 
     if (con->options.auth_plugin)
     {
-      snprintf((char *)ptr, DRIZZLE_MAX_SCRAMBLE_SIZE, "%s", con->password);
+      strncpy((char *)ptr, con->password, DRIZZLE_MAX_SCRAMBLE_SIZE);
```
ben-palmer-sociomantic, line 405: Just comment for the future, we might want some error handling here as that is a very short password: https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/597d2be70991be25c4545bf8031deb14612231e1/include/libdrizzle-redux/constants.h#L87

federico-cuello-sociomantic: It's scrambled. Why do you think it's very short?

ben-palmer-sociomantic: Ha! I just thought it was very short :)

PR #290, opened 2018-04-19
Remove validation of SSL_CTX_set_options

andreas-bok-sociomantic:
The call SSL_CTX_set_options to returns the new
options bitmask after adding options not 1 for
success


andreas-bok-sociomantic:
Wrong branch :(

PR #293, opened 2018-04-23
Avoid variable shadowing in unittests

andreas-bok-sociomantic:
Use `__ret` as variable name instead of ret to hold the return value for string comparison macros in `lite.h`

### DIFF
--- a/yatl/lite.h
+++ b/yatl/lite.h
@@ -227,8 +227,8 @@ do \
 { \
   size_t __expected_length; \
   size_t __actual_length; \
-  int ret= yatl_strcmp(__expected_str, __actual_str, &__expected_length, &__actual_length); \
-  if (ret) { \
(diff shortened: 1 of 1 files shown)


ben-palmer-sociomantic:
Where is `ret` declared that the old variable names were shadowing?

andreas-bok-sociomantic:
> Where is ret declared that the old variable names were shadowing?

The variable is defined in `lite.h` e.g. here [`lite.h`](https://github.com/sociomantic-tsunami/libdrizzle-redux/blob/v6.x.x/yatl/lite.h#L230)

Currently the drizzle return value is not declared as ret to avoid exactly to shadowing. But would rather use `ret` in the actual unittets instead of `__ret` or `driz_ret` (the case at the moment)

ben-palmer-sociomantic:
Ah ok, so its not shadowing at the moment but will in the future?

andreas-bok-sociomantic:
Yes.

PR #297, opened 2018-05-31
Add distribution specific container scripts

andreas-bok-sociomantic:
Reduce the complexity of the existing `container_entrypoint.sh` by moving the distribution specific parts into separate scripts.

Which script is copied to the container depends on the
**$DIST_NAME** environment variable

The `centos` specific script also adds `yum` repositories for additional version of the `clang` compiler

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,7 @@ services:
       CXX: $DOCKER_CXX
       CC: $DOCKER_CC
     build:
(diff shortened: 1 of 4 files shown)


andreas-bok-sociomantic:
Rebased on latest v6.x.x

ben-palmer-sociomantic:
LGTM

PR #300, opened 2018-06-12
Separate API for getting signed and unsigned numeric values

andreas-bok-sociomantic:
Refactor statement_param.cc and add a templated function
for retrieval of signed and unsigned integer types.

Add API function for extracting unsigned integer and long types.

### DIFF
--- a/include/libdrizzle-redux/statement.h
+++ b/include/libdrizzle-redux/statement.h
@@ -383,6 +383,19 @@ const char *drizzle_stmt_get_string(drizzle_stmt_st *stmt, uint16_t column_numbe
                                     size_t *len, drizzle_return_t *ret_ptr);
 
 /**
+ * Get the unsigned int value for a column of a fetched row
+ *
+ * @param stmt The prepared statement object
+ * @param column_number The column number to get (starting at 0)
+ * @param ret_ptr A pointer to a drizzle_return_t to store the return status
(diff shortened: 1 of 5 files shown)


include/libdrizzle-redux/statement.h
```diff
@@ -392,8 +405,22 @@ const char *drizzle_stmt_get_string(drizzle_stmt_st *stmt, uint16_t column_numbe
  * @return The int value
  */
 DRIZZLE_API
-uint32_t drizzle_stmt_get_int(drizzle_stmt_st *stmt, uint16_t column_number,
-                              drizzle_return_t *ret_ptr);
+int32_t drizzle_stmt_get_int(drizzle_stmt_st *stmt, uint16_t column_number,
```
ben-palmer-sociomantic, line 447: Hmm, so this is actually changing the API. In this case it can't go in a minor.

ben-palmer-sociomantic (commented):
One query I want to clear up before reviewing again.

include/libdrizzle-redux/statement.h
andreas-bok-sociomantic: Yes. Realized this as well. Though the current API is broken since you can't `set` an signed numeric value for a query parameter. Will create a  `v7.0.0` Milestone and `v7.x.x` branch

PR #306, opened 2018-06-22
Bugfix: Add check for stmt=NULL in drizzle_stmt_execute

andreas-bok-sociomantic:
Avoids a possible **SEGV** fault if the passed `drizzle_stmt_st` object is not initialized


ben-palmer-sociomantic (approved):
LGTM

PR #312, opened 2018-06-27
[TRAVIS] Disable conditional osx build

andreas-bok-sociomantic:
Startup times for OSX builds on travis are now
acceptable. Hence the condition is disabled
in the travis build configuration


ben-palmer-sociomantic:
LGTM

PR #322, opened 2018-11-10
Minor fixes in README

andreas-bok-sociomantic:
use relative path to compiling.rst

the travis build status for the minor branch v6.1.x


ben-palmer-sociomantic:
Why do we change the travis path? Won't that mean we need to keep changing it for every release?

andreas-bok-sociomantic:
> Why do we change the travis path? Won't that mean we need to keep changing it for every release?

Yes. But IIUC the travis build status wouldn't be correct otherwise. Will check the branch/tag which used if no qparams is passed to the travis build status svg

PR #326, opened 2018-12-02
Add const to params for non-mutating functions

andreas-bok-sociomantic:
Adds `const` qualifier to **POD** params which are not mutated by the function.

Since it is (possibly) a breaking change for clients it is added to the milestone for the next `major` release

### DIFF
--- a/include/libdrizzle-redux/binlog.h
+++ b/include/libdrizzle-redux/binlog.h
@@ -107,7 +107,7 @@ drizzle_return_t drizzle_binlog_start(drizzle_binlog_st *binlog,
 * @return The timestamp for the binlog event
 */
 DRIZZLE_API
-uint32_t drizzle_binlog_event_timestamp(drizzle_binlog_event_st *event);
+uint32_t drizzle_binlog_event_timestamp(const drizzle_binlog_event_st *event);
 
 /**
 * Get the event type for the event received by the event callback
@@ -116,7 +116,7 @@ uint32_t drizzle_binlog_event_timestamp(drizzle_binlog_event_st *event);
 * @return The event type of the binlog event
 */
 DRIZZLE_API
-drizzle_binlog_event_types_t drizzle_binlog_event_type(drizzle_binlog_event_st *event);
+drizzle_binlog_event_types_t drizzle_binlog_event_type(const drizzle_binlog_event_st *event);
 
 /**
 * Get the server_id for the event received by the event callback
@@ -125,7 +125,7 @@ drizzle_binlog_event_types_t drizzle_binlog_event_type(drizzle_binlog_event_st *
 * @return The server_id for the binlog event
 */
 DRIZZLE_API
-uint32_t drizzle_binlog_event_server_id(drizzle_binlog_event_st *event);
+uint32_t drizzle_binlog_event_server_id(const drizzle_binlog_event_st *event);
 
 /**
 * Get the length of the event data (excluding header) received by the event
@@ -135,7 +135,7 @@ uint32_t drizzle_binlog_event_server_id(drizzle_binlog_event_st *event);
 * @return The event data length
 */
 DRIZZLE_API
-uint32_t drizzle_binlog_event_length(drizzle_binlog_event_st *event);
+uint32_t drizzle_binlog_event_length(const drizzle_binlog_event_st *event);
 
 /**
 * Get the next event position from the event received by the event callback
@@ -144,7 +144,7 @@ uint32_t drizzle_binlog_event_length(drizzle_binlog_event_st *event);
 * @return The next event position
 */
 DRIZZLE_API
-uint32_t drizzle_binlog_event_next_pos(drizzle_binlog_event_st *event);
+uint32_t drizzle_binlog_event_next_pos(const drizzle_binlog_event_st *event);
 
 /**
 * Get the flags for the event received by the event callback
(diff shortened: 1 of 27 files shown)


include/libdrizzle-redux/binlog.h
```diff
@@ -200,7 +200,7 @@ uint32_t drizzle_binlog_event_raw_length(drizzle_binlog_event_st *event);
 * @return The event type of the binlog event as string
 */
 DRIZZLE_API
-const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_type);
+const char *drizzle_binlog_event_type_str(const drizzle_binlog_event_types_t event_type);
```
daniel-zullo, line 203: is it needed here?

include/libdrizzle-redux/binlog.h
```diff
@@ -231,7 +231,8 @@ const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_typ
  */
 DRIZZLE_API
 drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
-                                             uint32_t *end_position, int file_index);
+                                             uint32_t *end_position,
+                                             int file_index);
```
daniel-zullo: this change does not seem to be related

include/libdrizzle-redux/drizzle.h
```diff
@@ -106,7 +106,7 @@ const char *drizzle_bugreport(void);
  * @return String form of verbose level.
  */
 DRIZZLE_API
-const char *drizzle_verbose_name(drizzle_verbose_t verbose);
+const char *drizzle_verbose_name(const drizzle_verbose_t verbose);
```
daniel-zullo, line 109: why is it needed here?

include/libdrizzle-redux/return.h
```diff
@@ -88,7 +88,7 @@ typedef enum drizzle_return_t drizzle_return_t;
 * @param[in] ret result code
 * @return true on success, false otherwise
 */
-static inline bool drizzle_success(drizzle_return_t ret)
+static inline bool drizzle_success(const drizzle_return_t ret)
```
daniel-zullo, line 91: is it needed?

include/libdrizzle-redux/return.h
```diff
@@ -104,7 +104,7 @@ static inline bool drizzle_success(drizzle_return_t ret)
 * @param[in] ret result code
 * @return true on fail, false otherwise
 */
-static inline bool drizzle_failed(drizzle_return_t ret)
+static inline bool drizzle_failed(const drizzle_return_t ret)
```
daniel-zullo, line 107: needed ?

src/binlog.cc
```diff
@@ -443,7 +443,7 @@ drizzle_return_t drizzle_state_binlog_read(drizzle_st *con)
   return DRIZZLE_RETURN_OK;
 }
 
-const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_type)
+const char *drizzle_binlog_event_type_str(const drizzle_binlog_event_types_t event_type)
```
daniel-zullo, line 446: check this one too please

src/binlog.cc
```diff
@@ -490,7 +490,8 @@ const char *drizzle_binlog_event_type_str(drizzle_binlog_event_types_t event_typ
 }
 
 drizzle_return_t drizzle_binlog_get_filename(drizzle_st *con, char **filename,
-                                             uint32_t *end_position, int file_index)
+                                             uint32_t *end_position,
+                                             int file_index)
```
daniel-zullo: it does not seem to be related

src/column.cc
```diff
@@ -204,7 +204,7 @@ drizzle_column_type_t drizzle_column_type(drizzle_column_st *column)
   return column->type;
 }
 
-const char *drizzle_column_type_str(drizzle_column_type_t type)
+const char *drizzle_column_type_str(const drizzle_column_type_t type)
```
daniel-zullo, line 207: why?

daniel-zullo (commented):
I have a few questions and comments and rather than that LGTM

bokchan:
`const` was added to all parameters which weren't mutated in the function. However as you note it might be redundant for the arguments passed by value.

bokchan:
Seems there isn't a general agreement on whether to pass primitive types by `const`. It is recommended not to pass such as `const ref`, e.g. `(const int &i)`. 

- https://lefticus.gitbooks.io/cpp-best-practices/content/04-Considering_Safety.html
- https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters

Thus using `const` in cases here doesn't make much difference wrt. efficiency but does provide the clarity in the API that the function won't mutate the argument.

bokchan:

bokchan:
Documentation needs to be updated

andreas-bok-sociomantic:
Updated rst documentation

ben-palmer-sociomantic:
The `Use relnotes folder to describe new features` commit seems unrelated to this PR. It could easily be merged on its own.

bokchan:
> The Use relnotes folder to describe new features commit seems unrelated to this PR. It could easily be merged on its own.

Hmm true.  `v7.x.x` is behind `v6.x.x` but somehow I managed to get the commit with the relnotes folder onto it 🤦‍♂️ .  I will block this until `v6.2.0` has been released and `v6.x.x` has been merged into `v7.x.x`

PR #332, opened 2019-01-07
Use nullptr instead of NULL in library code

bokchan:
The library is compiled with the c++11 standard which recommends using nullptr instead of NULL


ben-palmer-sociomantic:
LGTM

ben-palmer-sociomantic:
Could this just go in v6.2.0?

bokchan:
> Could this just go in v6.2.0?

Yes it could. Planning to make v7.0 use more c++11 features/guidelines, so would rather wait and bundle it with the next major release

PR #342, opened 2019-10-14
Fix centos build

daniel-zullo:
The clang package was updated and the previous version no longer exist
in the fedora project repository.


daniel-zullo:
Self-rebasing as it is a trivial fix

